PR #251, opened 2014-12-22
Add check for max length of strings, reduce classes size

flodolo:
**LangManager**
- Extract MAX_LENGTH from lang files
- Move ['python_vars'] to ['errors']['python'] to support more types of errors
- Add function countErrors
- Move functions for .po files into a new class
- Move error checks in separate functions (python, length)

**Utils**
- starsWith: accept array of strings as needle
- Use mb_strlen instead of strlen, add getLength

**GetTextManager**
- Functions to manage .po files originally in LangManager

**Views**
Expose number of errors in JSON, display errors in listsitesperlocale and errors view

Clean up phpDOC syntax in all classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 config/settings.inc.php
+composer.phar
 composer.lock
 locales
 vendor/
--- a/classes/Langchecker/DotLangParser.php
+++ b/classes/Langchecker/DotLangParser.php
@@ -1,7 +1,7 @@
 <?php
 namespace Langchecker;
 
-/*
+/**
  * DotLangParser class
  *
  * This class is for all the methods we use to read .lang files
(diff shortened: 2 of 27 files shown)


pascalchevrel:
Maybe we should maintain a separate repo for the libs we use across projects and include them via composer, I checked Composer documentation and it seems possible to include packages from repositories not registered in Packagist.

flodolo:
Definitely worth it (probably not in this PR). We now have a Bugzilla class across langchecker+webdashboard, we could probably build a class for generic string related functions, we already have Json (in completely different statuses) in Transvision+langchecker+webdashboard.

flodolo:
Did some tests and it's indeed not too difficult to do, but I think it's easier to discuss it in a different PR
https://github.com/flodolo/php-bugzilla-class

pascalchevrel:
Tested locally, works fine, r+ with nits

flodolo:
Thanks. Fixed nits and squashed, I'll wait for the webdashboard's one to merge.

flodolo:
Actually leaving the merge to you since this requires a ```php composer.phar update``` on the server to avoid exploding.

PR #254, opened 2014-12-24
Remove critical status for Thunderbird page

flodolo:
I don't think that Thunderbird pages should have critical status, but I'm not the one to make the call ;-)
Leaving the PR here for now.

### DIFF
--- a/config/sources.inc.php
+++ b/config/sources.inc.php
@@ -153,7 +153,6 @@ $lang_flags['www.mozilla.org'] = [
     'newsletter.lang'                         => [
         'critical' => ['de', 'es-ES', 'fr', 'hu', 'id', 'pl', 'pt-BR', 'ru'],
     ],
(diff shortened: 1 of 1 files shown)


pascalchevrel:
The start page promotes the 10 year anniversary so it definitely has a deadline. It is also a key in-product page for Thunderbird, so it should have the same priority as an in-product page in Firefox, it is also today the only existing communication channel that the Thunderbird project has with their users. I think it is enough reasons to mark it as critical.

flodolo:
I'm just afraid that we're sending out mixed messages: the page is marked as critical (as in "critical in itself", because that's the only status we track on the webdashboard), but we're also telling people to ignore it if they don't have resources or don't plan to work on Thunderbird pages.

Should we plan to start displaying extra tags on the webdashboard? Anyhow, I'm fine in closing this one and leave things as they are.

PR #258, opened 2015-01-06
Reorganize sources and list of locales

flodolo:
* Remove $mozilla_europe array (unused).
* Use $mozilla as list of all locales somehow supported by langchecker (but not necessarily by mozilla.org). Remove from this array all dropped locales.
* Add a list of locales supported on Firefox Desktop (e.g. tiles, FHR).
* Use $mozillaorg as list of locales supported by mozilla.org. Use it for mozilla.org pages instead of $mozilla
* Move other arrays of locales in sources, use the same style and order them alphabetically (no changes to locales for this part)

Also don't display empty website in listsitesforlocale, stopping at the beginning if they're not supported.

### DIFF
--- a/config/locales.inc.php
+++ b/config/locales.inc.php
@@ -1,47 +1,28 @@
 <?php
 
-$mozilla_europe = [
-    'bg', 'ca', 'cs', 'da', 'de', 'el', 'en-GB', 'es-ES', 'eu', 'fi',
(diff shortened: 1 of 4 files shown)


flodolo:
@pascalchevrel Mind taking a look before merge? It should be pretty safe, but I've already managed to mess a one-line change today.

I'm ignoring warnings from Bugzilla (missing 'es' and 'sr-Latn'), since I plan to fix them properly later this week (externalize the class).

pascalchevrel:
I am not seeing anything wrong, merging

PR #260, opened 2015-01-08
Remove Bugzilla class, use external lib for Bugzilla

flodolo:
Langchecker doesn't have a cache folder at the moment, so we need to make sure it's writable by PHP on the community server.


composer.json
```diff
@@ -1,17 +1,29 @@
...
-        "atoum/atoum" : "dev-master",
+        "atoum/atoum": "dev-master",
+        "mozillal10n/bugzilla": "0.*",
+        "mozillal10n/cache": "0.*",
         "sepia/po-parser": "3.0.5"
     },
 
     "require-dev": {
-        "raveren/kint" : "dev-1.0.0-wip"
+        "raveren/kint": "dev-1.0.0-wip"
```
pascalchevrel: we could take this opportunity tu use var_dumper instead of kint like on Transvision

flodolo: I agree. I'll replicate what you did there.

flodolo:
Updated both PR with Symfony/VarDumper and the same helper from Transvision to have ```cli_dump()```

PR #264, opened 2015-01-15
turn en-GB into a normal locale, use en-US as reference

pascalchevrel:


flodolo:
Nit: we need to fix also ```/scripts/import_transvision_id```, the rest looks good.

Issue #15, opened 2013-09-28
Import locamotion data


pascalchevrel:
We have no control over what is on locamotion, they import our lang files into their po based system, some people translate (few) and sometimes Dwayne sends the lang files back to svn, but this is once every 10 days at best and many times our deadlines are passed or we decide to redirect a minor locale page because we just don't know if they translated it.

I found export folders on locamotion with po files in them, it seems to be updated regularly. The branch locamotion_import allows to import all that strings when we use lang_update.

syntax:
./lang_update firefox/new.lang 0 all locamotion

since this adds a 4th parameter, the default values of the optional ones have to be indicated.

It seems to work well, I couldn't find a regression on all the features langchecker offers. One drawback is that it is very slow as i makes a request per locale to locamotion but it seems to be a good measure to not loose translations from minor locales.

PR #273, opened 2015-02-02
Create a view to display opt-in pages for a locale

flodolo:
Opt-in status is determined by a `opt-in` flag associated with the file.

### DIFF
--- a/views/optin.inc.php
+++ b/views/optin.inc.php
@@ -1,0 +1,88 @@
+<?php
+namespace Langchecker;
+
+use \Bugzilla\Bugzilla;
+
+$html_output = '<p id="back"><a href="http://l10n.mozilla-community.org/webdashboard/">Back to Web Dashboard</a></p>';
+
+// This view works only for mozilla.org (website ID 0)
+$current_website = $sites[0];
+$current_locale = $locale;
+
+if (! Project::isSupportedLocale($current_website, $current_locale)) {
+    exit($html_output . '<p>This locale is not supported on mozilla.org</p>');
+}
+
+// Create a list of opt-in pages
+$optin_pages = [];
+foreach (Project::getWebsiteFiles($current_website) as $current_filename) {
+    if (in_array('opt-in', Project::getFileFlags($current_website, $current_filename, $current_locale))) {
+        $optin_pages[$current_filename] = Project::getSupportedLocales($current_website, $current_filename, $langfiles_subsets);
+    }
+}
+
+if (count($optin_pages) == 0) {
+    exit($html_output . '<p>There are no optional pages available at the moment.</p>');
+}
+
+$html_output .= "<h1>List of optional pages for <span>{$current_locale}</span></h1>
+  <table class='optinpages'>
+    <thead>
+      <tr>
+        <th>Filename</th>
+        <th>Strings</th>
+        <th>Words</th>
+        <th>Opted-in</th>
+        <th>Actions</th>
+      </tr>
+    </thead>
+    <tbody>\n";
+
+$bugzilla_locale_name = urlencode(Bugzilla::getBugzillaLocaleField($current_locale, 'www'));
+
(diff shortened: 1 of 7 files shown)


views/optin.inc.php
```diff
@@ -0,0 +1,92 @@
...
+                       . '&flag_type-418=X&flag_type-419=X&flag_type-506=X&flag_type-507=X&form_name=enter_bug'
+                       . '&keywords=&maketemplate=Remember%20values%20as%20bookmarkable%20template&op_sys='
+                       . 'All&priority=--&product=www.mozilla.org&qa_contact=pascalc%40gmail.com'
+                       . '&rep_platform=All&short_desc=%5Bl10n%3A ' . $current_locale . '%5D%20Add%20%27'
+                       . $current_locale . '%27%20to%20the%20list%20of%20supported%20locales%20for%20'
+                       . $current_filename .'&target_milestone=---&version=unspecified'
+                       . '&format=__default__&cf_locale=' . $bugzilla_locale_name;
+
+        $status = '<span class=\'no\'>no</span> ';
+        $actions = "<a href='{$bugzilla_link}' title='File a bug to request this page'>Opt-in</a>";
```
pascalchevrel: You probably need to add the version flag to 'Development/Staging' otherwise the bug can't be filed directly on Bugzilla

views/optin.inc.php
```diff
@@ -0,0 +1,92 @@
+<?php
+namespace Langchecker;
+
+use \Bugzilla\Bugzilla;
+
+$htmloutput = '<p id="back"><a href="http://l10n.mozilla-community.org/webdashboard/">Back to Web Dashboard</a></p>';
```
pascalchevrel: $html_output (I know we have other views where we use $htmloutput but let's stick to conventions for new content :) )

views/optin.inc.php
```diff
@@ -0,0 +1,92 @@
...
+
+// This view works only for mozilla.org (website ID 0)
+$current_website = $sites[0];
+$current_locale = $locale;
+
+if (! Project::isSupportedLocale($current_website, $current_locale)) {
+    $htmloutput .= "<p>This locale is not supported on mozilla.org</p>";
+    echo $htmloutput;
+    exit;
```
pascalchevrel: These 3 lines could just be:
```php
exit($htmloutput . '<p>This locale is not supported on mozilla.org</p>');
```
(also, single quotes when there are no variable expansions)

views/optin.inc.php
```diff
@@ -0,0 +1,92 @@
...
+foreach (Project::getWebsiteFiles($current_website) as $current_filename) {
+    if (in_array('opt-in', Project::getFileFlags($current_website, $current_filename, $current_locale))) {
+        $optin_pages[$current_filename] = Project::getSupportedLocales($current_website, $current_filename, $langfiles_subsets);
+    }
+}
+
+if (count($optin_pages) == 0) {
+    $htmloutput .= "<p>There are no optional pages available at the moment.</p>";
+    echo $htmloutput;
+    exit;
```
pascalchevrel: same comment as line #15

flodolo:
You're right about Bugzilla URL, I also fixed the place the other template where I copied it from

pascalchevrel:
Ideally, it would be cool to put links to the real page on mozilla.org next to the lang file, so as that localizers know for sure what page they are opting in for.

flodolo:
I thought about it, also for other pages (especially those moving around on demo servers, not www-dev initially). 

My idea would be to have a tag similar to ```## NOTE:```in the reference file (e.g. ```## DEMOURL:```), that doesn't propagate in localized files like ```## TAG:``` or ´´´## MAX_LENGTH:```,  to have a reliable way to determine the page URL, but that requires a bit of work.

pascalchevrel:
we can definitely see the links later

PR #279, opened 2015-02-13
Integrate Travis, fix missing 'static' causing the build to fail

flodolo:


flodolo:
@pascalchevrel I was thinking: would it make sense to have a #l10n-tools channel on IRC and use it for all Travis+Github notifications?

P.S. I tried to copy Transvision IRC section in travis.yml, but it didn't seem to work

pascalchevrel:
The problem with centralizing notifications in a separate channel is that those notifications would become invisible to Transvision contributors, so #l10n-tools sounds good to me for our team needs as long as as project like Transvision which is more open to external contributions is not included.

pascalchevrel:
Also, the IRC part needs settings to be added in the preferences for the repo

flodolo:
I wasn't clear: I don't want to move Transvision's notifications, those belong to #transvision.

I was talking of all the tools we have here in mozilla-l10

pascalchevrel:
then yes, that's probably a good idea

pascalchevrel:
Then yes, all notifications to our internal tools into a separate IRC channel is fine, we can also talk there about the tools intead of the  l10n-drivers channel

PR #284, opened 2015-02-18
Assign Firefox Desltop pages to locales that do localize Firefox Desktop...

pascalchevrel:
..., not all locales

### DIFF
--- a/config/sources.inc.php
+++ b/config/sources.inc.php
@@ -715,12 +715,12 @@ $langfiles_subsets = [
                 'nl', 'pl', 'pt-PT', 'ro', 'sk', 'sl', 'sv-SE',
             ],
         'firefox/android/index.lang'            => $android_locales,
-        'firefox/australis/firefox_tour.lang'   => $mozillaorg,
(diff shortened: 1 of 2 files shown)


config/sources.inc.php
```diff
@@ -715,12 +715,12 @@
...
         'firefox/android/index.lang'            => $android_locales,
-        'firefox/australis/firefox_tour.lang'   => $mozillaorg,
-        'firefox/australis/fx36_tour.lang'      => $mozillaorg,
-        'firefox/channel.lang'                  => $mozillaorg,
-        'firefox/desktop/customize.lang'        => $mozillaorg,
-        'firefox/desktop/fast.lang'             => $mozillaorg,
-        'firefox/desktop/index.lang'            => $mozillaorg,
+        'firefox/australis/firefox_tour.lang'   => $firefox_desktop,
+        'firefox/australis/fx36_tour.lang'      => $firefox_desktop,
+        'firefox/channel.lang'                  => $firefox_desktop,
```
flodolo: We should probably keep this one as `$mozillaorg´, since it has download buttons for Firefox for Android too. I agree on all the other pages.

PR #288, opened 2015-02-24
Add view to get all translated snippets for a locale in JSON format

flodolo:


flodolo:
The idea is to have a JSON with all translated snippets for a locale. Hopefully I'll convince the snippets team to import a localization from this API instead of doing manual copy and paste

views/snippets_api.inc.php
```diff
@@ -0,0 +1,44 @@
...
+
+use Transvision\Json;
+
+// This view works only for snippets (website ID 6)
+$current_website = $sites[6];
+$current_locale = $locale;
+$json_data = [];
+
+if ($locale == '') {
+    exit('<p>Missing locale code in the request</p>');
```
pascalchevrel: We should return a 400 error with the error message, as we do on Transvision, maybe add this static method to the Json class:
```php
public static function invalidAPICall($error)
{
    http_response_code(400);
    return self::output(['error' => $error], false, true)
}
```

views/snippets_api.inc.php
```diff
@@ -0,0 +1,44 @@
...
+}
+
+if (! Project::isSupportedLocale($current_website, $current_locale)) {
+    exit('<p>This locale is not supported for snippets</p>');
```
pascalchevrel: same as above

views/snippets_api.inc.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+ksort($json_data);
+
+if (isset($_GET['callback'])) {
+    echo Json::output($json_data, $_GET['callback'], true);
+} else {
+    echo Json::output($json_data, false, true);
+}
```
pascalchevrel: Simpler I think:
```php
$callback = isset($_GET['callback']) ? $_GET['callback'] : false;
exit(Json::output($json_data, $callback, true));
```
Generally speaking, I prefer to exit when we output json, because we know we have finished processing the request and any other data output afterwards would be a bug.

pascalchevrel:
Generally speaking this is ok, note though that all out strings are already available via an existing API per snippet and not per locale, ex:
https://l10n.mozilla-community.org/~pascalc/langchecker/?action=api&file=mar2015_a.lang
https://l10n.mozilla-community.org/~pascalc/langchecker/?action=api&file=mar2015_a.lang&stringid=bab38705dcf177f21d7fe25150e4610e943992e2

This json source was created only for the consumption of the marketing team in charge of snippets so as that we could automate publishing of snippets but they refused to use it and preferred to create a new web app, so not sure that adding an alternate API will make them change their mind.

PR #301, opened 2015-03-27
update Langchecker to use ADI data located on the server

pascalchevrel:


views/getcoverage.inc.php
```diff
@@ -1,7 +1,13 @@
...
 namespace Langchecker;
 
-include __DIR__ . '/../config/adu.inc.php';
+// Real data is in adi.inc.php, not under VCS
+if (is_file(__DIR__ . '/../config/adi.inc.php')) {
+    include __DIR__ . '/../config/adi.inc.php';
+} else {
+    // Fake data to not break the app outside of production
+    include __DIR__ . '/../config/fake_adi.inc.php';
+}
```
flodolo: Uhm... Do we actually need this? We should already have access to ´´´sources.php´´´

flodolo:
Besides the note on `getcoverage.php`, my only suggestion would be to have a `archive_adi.inc.php` instead of a `fake_adi.inc.php`, and use the content currently in `adu.inc.php´.
At least data would still make some sort of sense. But I can also live with the current proposed structure.

views/getcoverage.inc.php
pascalchevrel: You're right, we don't need it and didn't need it before either, removing

PR #305, opened 2015-04-21
Switch Transvision scripts to SSL, support locale mapping for Pootle

flodolo:


flodolo:
Merging since it's needed for pt-PT import.

Issue #312, opened 2015-05-05
Improve perf on errors page


pascalchevrel:
I have patches that improve significantly the performance of our error page which can take more than 30s to be processes. That actually improves all views as it makes our parser faster but this specific page is the one where this is most visible.

PR #313, opened 2015-05-06
Issue #312: improve performance of error page

pascalchevrel:
- faster Utils::isUTF8() by giving the path of the Magic mime file on the server
- store parsed English reference files in a static property instead of reparsing them every time
- in DotLangParser::ParseFile() avoid reparsing the whole file in the main loop for file metadata once we are already parsing strings
- Utils::starstWith() && Utils::leftStrip() have specific cases for searching ; and # with less calculation (on the error page we call these functions up to 2 million times)
- new DotLangParser::getMetaTags() utility function to return the list of tags we use in lang files
- moved parsing of reference metadata to a private DotLangParser::extractReferenceMetaData() property to make parseFile() more readable

Locally, the new version is 4 times faster for me, since this page is heavy on CPU use, results may be lower on our community server which is not particularily fast.

### DIFF
--- a/classes/Langchecker/DotLangParser.php
+++ b/classes/Langchecker/DotLangParser.php
@@ -12,6 +12,14 @@ namespace Langchecker;
 class DotLangParser
 {
     /**
+     * We store in this variable the result of lang files already parsed because
+     * parsing is CPU intensive
+     *
+     * @var array
+     */
+    private static $parsed_files = [];
+
+    /**
      * Load file, remove empty lines and return an array of strings
      *
      * @param string  $path        Filename to analyze
@@ -51,19 +59,20 @@ class DotLangParser
      */
     public static function parseFile($path, $reference_locale = false)
     {
-        $dotlang_data = [];
-        // In case the file is missing we set an empty array for strings
-        $dotlang_data['strings'] = [];
+        // Check if we haven't already parsed the file
+        if (isset(self::$parsed_files[$path])) {
+            return self::$parsed_files[$path];
+        }
+
         $file_content = self::getFile($path);
 
-        // All meta tags in format “## METATAG:”, but not file tags (“## TAGNAME ##”)
-        $meta_tags = ['NOTE', 'TAG', 'MAX_LENGTH', 'URL'];
-        $meta_tags = array_map(
-                        function ($tag) {
-                            return "## {$tag}:";
-                        },
-                        $meta_tags
-                     );
+        $dotlang_data = [];
+
+        // In case the file is missing we set an empty array for strings
+        $dotlang_data['strings'] = [];
+
+        // Make sure there's always an activation status to avoid controls later
+        $dotlang_data['activated'] = false;
 
         if ($file_content !== false) {
             // First pass: extract strings and metadata (tags, active status) relevant for all locales.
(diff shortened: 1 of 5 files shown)


classes/Langchecker/DotLangParser.php
```diff
@@ -12,6 +12,13 @@
 class DotLangParser
 {
     /**
+     * We store in this variable the result of lang files already parsed because
+     * parsing is CPU intensive
+     * @var array
+     */
+    public static $parsed_files = [];
+
```
flodolo: I'd leave an empty line before ```@var```. We don't need ```$parsed_files``` to be public.

classes/Langchecker/DotLangParser.php
```diff
@@ -78,20 +86,25 @@ function ($tag) {
                 }
 
                 // Get file description
-                if (Utils::startsWith($current_line, '## NOTE:')) {
+                if (! isset($dotlang_data['filedescription'])
```
flodolo: We support multiple ```## NOTE:##``` to avoid having huge lines of text (example in screenshots), so we shouldn't check if ```filedescription``` is already set.

classes/Langchecker/DotLangParser.php
```diff
@@ -122,57 +133,88 @@ function ($tag) {
...
+            }
+        }
+
+        return $dotlang_data;
+    }
+
+    /**
+     * Provide all meta tags in format “## METATAG:”,
+     * but not file tags (“## TAGNAME ##”)
+     *
```
flodolo: Leave only the first line of the comment, the second one is not relevant for the function.

classes/Langchecker/DotLangParser.php
```diff
@@ -122,57 +133,88 @@ function ($tag) {
...
+        );
+
+        return $meta_tags;
+    }
+
+    /**
+     *  Extract the metadata in English reference file
+     *
+     * @param  array $file_content Content of a lang file
+     * @return array Parsed lang file with the associated reference metadata
```
flodolo: Empty line before `@return` for consistency.

classes/Langchecker/Utils.php
```diff
@@ -21,6 +21,11 @@ class Utils
      */
     public static function leftStrip($origin, $substring)
     {
+        // Lang file common cases: reference string or comment
+        if ($substring === ';' || $substring === '#') {
+            return trim(substr($origin, 1));
```
flodolo: Should we use `mb_substr` instead?

templates/template.inc.php
```diff
@@ -17,8 +17,12 @@
...
 <?php
-    echo "<!-- Elapsed time (s): " . round((microtime(true) - $_SERVER["REQUEST_TIME_FLOAT"]), 4) . " -->\n";
-    echo "<!-- Memory usage (MB): " . round(memory_get_peak_usage(true) / (1024 * 1024), 2) . " -->\n";
+    $time   = round((microtime(true) - $_SERVER["REQUEST_TIME_FLOAT"]), 4);
+    $memory = round(memory_get_peak_usage(true) / (1024 * 1024), 2);
+    echo "<!-- Elapsed time (s): " . $time . " -->\n";
+    echo "<!-- Memory usage (MB): " . $memory . " -->\n";
+    error_log('Elapsed time (s):'  . $time);
+    error_log("Memory usage (MB): " . $memory);
 ?>
```
flodolo: I don't think we should write this in the error_log, it creates a lot of noise and make real errors harder to spot.

flodolo:
Nice: time on my machine is 4x faster, memory increases about 3x but it's still very low. The only problem is with the file description, rest is just minor nits.

In general I wonder if we should move to multi byte functions when touching code (e.g. ```strpos``` -> ```mb_strpos```)

PR #318, opened 2015-05-19
More performance improvement, mostly on the action=count view:

pascalchevrel:
- DotLangParser::extract_metadata static boolean property, allows a view to not extract any metadata when parsing lang files
- LangManager::error_checking static boolean property, allows a view to not look for errors in strings when parsing lang files
- Update main loop in DotLangParser::ParseFile() to test most common case first (translation exists)
- many small nitpicks and cleanups

Performance gains are mostly on the 'count' view (40% for me) but other views also benefit from a small speedup because of the code reorganization (~10%).

### DIFF
--- a/classes/Langchecker/DotLangParser.php
+++ b/classes/Langchecker/DotLangParser.php
@@ -20,19 +20,31 @@ class DotLangParser
     private static $parsed_files = [];
 
     /**
+     * We store in this variable if we want to extract file metadata for a view.
+     *
+     * @var boolean
+     */
+    public static $extract_metadata = true;
+
+    /**
+     * Logging errors has a performance impact avoid on public views
+     *
+     * @var boolean
+     */
+    public static $log_errors = true;
+
+    /**
      * Load file, remove empty lines and return an array of strings
      *
-     * @param string  $path        Filename to analyze
-     * @param boolean $show_errors Display or not errors in case of
-     *                             missing file
+     * @param string $path Filename to analyze
      *
      * @return mixed Cleaned up array of lines (array),
      *               or false if file is missing
      */
-    public static function getFile($path, $show_errors = true)
+    public static function getFile($path)
     {
         if (! is_file($path)) {
-            if ($show_errors) {
+            if (self::$log_errors) {
                 Utils::logger("{$path} does not exist.");
             }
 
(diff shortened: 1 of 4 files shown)


pascalchevrel:
Failing test, didn't notice, will update the PR soon

classes/Langchecker/DotLangParser.php
```diff
@@ -29,10 +43,10 @@ class DotLangParser
      * @return mixed Cleaned up array of lines (array),
```
flodolo: Remove show_errors from method doc.

classes/Langchecker/DotLangParser.php
```diff
@@ -127,24 +145,31 @@ public static function parseFile($path, $reference_locale = false)
...
                     }
 
-                    // Store the translation
-                    $dotlang_data['strings'][$reference] = $translation;
-
-                    $i++;
+                    /* Empty translation: what I'm reading as translation is either the next reference string
+                     * or the next meta tag (comment, tag binding). I consider this string untranslated.
+                     */
+                    $dotlang_data['strings'][$reference] = $reference;
```
flodolo: Completely missed this in the previous update :-\

classes/Langchecker/DotLangParser.php
```diff
@@ -127,24 +145,31 @@ public static function parseFile($path, $reference_locale = false)
...
-                        $dotlang_data['strings'][$reference] = $reference;
+                    // If we don't extract metadata, we shouldn't test for #
+                    $markers = self::$extract_metadata
+                        ? [';', '#']
+                        : [';'];
+
+                    // We test the most common scenario first: reference string followed by a translation
+                    if (! Utils::startsWith($translation, $markers)) {
+                        // Store the translation
+                        $dotlang_data['strings'][$reference] = $translation;
```
flodolo: Should we set `$translation` in this conditional, considering that it's now used only here?

flodolo:
Besides the comment, looks good to me.

classes/Langchecker/DotLangParser.php
pascalchevrel: it needs to be set before the conditionnal because the if() close makes a test on the $translation variable

flodolo: Right, missed the previous line.

PR #327, opened 2015-05-27
Improve management of obsolete files

flodolo:
* Add Project::isObsoleteFile method + tests
* Don’t display them in the activation view
* Display a warning in global status header when a file is obsolete
* Ignore obsolete files in mark_active


pascalchevrel:
LGTM

PR #337, opened 2015-07-01
Add functional tests to Langchecker JSON API

TheoChevalier:


TheoChevalier:
Since the check function is the same as in Transvision, maybe we could add it to the Verif library?

TheoChevalier:
Not sure what to do about that:
-------------------------------------
Report for: Check API HTTP responses
-------------------------------------
Error:
http://localhost:8083/api/?website=foo&file=main.lang&json is not valid Json
http://localhost:8083/api/?action=coverage&locales[]=it&locales[]=fr is not valid Json
http://localhost:8083/api/?action=coverage&locales[]=fr is not valid Json
http://localhost:8083/api/?action=coverage&locales[]=foo is not valid Json
http://localhost:8083/api/?action=coverage&locales=foo is not valid Json
http://localhost:8083/api/?action=coverage is not valid Json
http://localhost:8083/api/?action=api&file=euballot.lang is not valid Json

flodolo:
Try running the file locally, you'll notice you're trying to get `api/?action etc.`instead of `/?action etc.`

tests/functional/api.php
```diff
@@ -0,0 +1,57 @@
...
+    ['?action=coverage&locales[]=foo', 200, '0.00'],
+    ['?action=coverage&locales=foo', 200, 'ERROR: locales is not an array.'],
+    ['?action=coverage', 200, 'ERROR: missing list of locales.'],
+    ['?action=api&file=euballot.lang', 200, "<ul>\n<li><a href='?action=api&file=euballot.lang&stringid=e1399441a00827af6b458672fa05d48462659638'>Firefox is fast, flexible and secure. Proudly non-profit, we're always innovating for you. Try the free download!</a></li>\n</ul>\n"],
+];
+
+$obj = new \pchevrel\Verif('Check API HTTP responses');
+$obj
+    ->setHost('localhost:8083')
+    ->setPathPrefix('api/');
```
flodolo: Don't need this

tests/functional/api.php
```diff
@@ -0,0 +1,57 @@
...
+// We will need the pid to kill it, beware, this is the pid of the bash process started above
+$pid = $output[0];
+
+// Pause to let time for the dev server to launch in the background
+sleep(3);
+
+$paths = [
+    ['?action=listlocales&json', 400, "[\n    \"Please check you request: provide a project name, or a valid couple website+file.\"\n]"],
+    ['?website=0&file=firefox/desktop/tips.lang&json&locale=fr', 200, "{\n    \"firefox\/desktop\/tips.lang\": {\n        \"fr\": {\n            \"Identical\": 0,\n            \"Translated\": 40,\n            \"Missing\": 0,\n            \"Obsolete\": 0,\n            \"tags\": [\n                \"sync_device_note\"\n            ],\n            \"activated\": true\n        }\n    }\n}"],
+    ['?website=foo&file=main.lang&json', 400, 'foo is not a supported website. Check the value and try again.'],
```
flodolo: Need a real website here, not `foo`

flodolo: Actually, never mind, you're testing for the error. But the output is not a JSON, is a simple string.

TheoChevalier: ok, so we have lots of results that are not JSON :/

TheoChevalier:
Fixed the '/api' (d'oh). But rresults are not JSON either.

tests/functional/api.php
flodolo: Yes, we have several of them. Either you duplicate `$check` and avoid doing `->isJson()` for these calls /ì(haven't tried but it should work), or update the code to always return a JSON.

TheoChevalier:
Ok, I think it's ready for review. I should have covered all the URL returning JSON or a string listed on https://github.com/mozilla-l10n/langchecker/wiki/Available-URLs

tests/functional/api.php
```diff
@@ -0,0 +1,60 @@
...
+// Make sure we have a timezone set
+date_default_timezone_set('Europe/Paris');
+
+require __DIR__ . '/../../vendor/autoload.php';
+
+// Launch PHP dev server in the background
+chdir(INSTALL_ROOT);
+exec('php -S 0.0.0.0:8083 > /dev/null 2>&1 & echo $!', $output);
+
+// We will need the pid to kill it, beware, this is the pid of the bash process started above
```
pascalchevrel: We are not launching a bash script but directly a php script, update the comment

tests/functional/api.php
```diff
@@ -0,0 +1,60 @@
...
+            ->isEqualTo($content);
+    }
+};
+
+$check($obj, $paths);
+
+$obj->report();
+
+// Kill PHP dev server by killing all children processes of the bash process we opened in the background
+exec('pkill -P ' . $pid);
```
pascalchevrel: This is not going to work here, we haven't launched a bash process (and actually, this is probably also wrong on Transvision). Killing the pid like this should work:
```php
exec('kill -1 ' . $pid);
```

inc/init.php
```diff
@@ -17,6 +17,11 @@
 require $conf . 'locales.inc.php';
 require $conf . 'sources.inc.php';
 
+// Override sources for functional tests
+if (getenv('TRAVIS')) {
```
pascalchevrel: That seems a bad approach, we need to get these tests to also work locally, not just on Travis. Generally speaking, testing for the structure of the data returned instead of the exact content which varies over time seems more reliable.

tests/functional/api.php
```diff
@@ -0,0 +1,60 @@
...
+    ['?action=snippets&locale=it', 200, "{\n    \"@@this is a test, do not remove@@\": \"Questo \u00e8 un test\",\n    \"Email\": \"Email\",\n    \"Fifth string with %(one)s, %(two)s\": \"Quinta stringa con %(one)s, %(two)s e ancora %(two)s\",\n    \"Fourth string with 90%% coverage\": \"Quarta stringa con copertura al 90%\",\n    \"Hello\": \"Ciao\",\n    \"Obsolete string\": \"Stringa obsoleta\",\n    \"Press center\": \"Sala stampa\",\n    \"Save file\": \"Salva file troppo lungo\",\n    \"Second string with %(num)s tags\": \"Seconda stringa con %(num)s etichette\",\n    \"Sixth string with %(one)s\": \"Sesta stringa con %(one)s e %(two)s\",\n    \"String with %(num)s tags\": \"Stringa con etichette e variabile sbagliata\",\n    \"String with multiple comments\": \"String con pi\u00f9 commenti\",\n    \"Third string with %s tags\": \"Terza stringa con variabile mancante\"\n}"],
+];
+
+$obj = new \pchevrel\Verif('Check API HTTP responses');
+$obj
+    ->setHost('localhost:8083');
```
pascalchevrel: keep on same line as $obj since we have only one method call.

tests/functional/api.php
```diff
@@ -0,0 +1,60 @@
...
+
+$check = function ($object, $paths) {
+    foreach ($paths as $values) {
+        list($path, $http_code, $content) = $values;
+        $object
+            ->setPath($path)
+            ->fetchContent()
+            ->hasResponseCode($http_code)
+            ->isEqualTo($content);
+    }
```
pascalchevrel: Since you removed the test on json feeds, you are not actually testing that we are sending valid json. You should separate your testing of pure text and json feeds or put in $paths what data you expect and add a conditional in the foreach loop to test for json or not.

flodolo:
For your happiness, added a new URL (this one returns a JSON also for errors)
https://l10n.mozilla-community.org/~pascalc/langchecker/?action=translate&file=careers.lang&json
https://l10n.mozilla-community.org/~pascalc/langchecker/?action=translate&file=foo.lang&json

TheoChevalier:
This now requires an update from Verif lib: https://github.com/pascalchevrel/Verif/pull/1
I've put an arbitrary 0.2 version number in composer.json

tests/functional/api.php
```diff
@@ -0,0 +1,93 @@
...
+};
+
+$check($obj, $json, 'json');
+$check($obj, $text, 'text');
+$check($obj, $number, 'number');
+
+$obj->report();
+
+// Kill PHP dev server we launched in the background
+exec('pkill -P ' . $pid);
```
pascalchevrel: Use ```kill -9 ``` or the process won't die, ```pkill -P``` is to kill the descendant processes (when we launch a shell that itself launches a php server for example), in this script, we launch the php process directly, not via a shell script.

tests/functional/api.php
```diff
@@ -0,0 +1,93 @@
...
+$number = [
+    ['?action=coverage&locales[]=it&locales[]=fr', 200, '1.92'],
+    ['?action=coverage&locales[]=fr', 200, '0.32'],
+    ['?action=coverage&locales[]=foo', 200, '0.00'],
+];
+
+$obj = new \pchevrel\Verif('Check API HTTP responses');
+$obj->setHost('localhost:8083');
+
+$check = function ($object, $paths, $type) {
```
pascalchevrel: You could do :
``` php
$check = function ($object, $paths, $type) use ($obj) {
```
and later:
```php
$check($json, 'json');
```

tests/functional/api.php
```diff
@@ -0,0 +1,93 @@
...
+                        ->isNumeric()
+                        ->isEqualTo($content);
+                break;
+                default:
+                    $object
+                        ->setPath($path)
+                        ->fetchContent()
+                        ->hasResponseCode($http_code)
+                        ->isEqualTo($content);
+                break;
```
pascalchevrel: There is a lot of method duplication, this would be probably simpler:
```php
$check = function ($paths, $type) use ($obj) {
        foreach ($paths as $values) {
            list($path, $http_code, $content) = $values;
            $obj
                ->setPath($path)
                ->fetchContent()
                ->hasResponseCode($http_code)
                ->isEqualTo($content);

            switch ($type) {
                case 'json':
                    $obj->isJson();
                    break;
                case 'number':
                    $obj->isNumeric();
                    break;
                default:
                    break;
            }
        }
};
```
Also please indent break statements:
https://github.com/php-fig/fig-standards/blob/master/accepted/fr/PSR-2-coding-style-guide.md#52-switch-case

PR #339, opened 2015-07-03
Add JSON output to Translate view

flodolo:


views/translatestrings.inc.php
```diff
@@ -31,7 +36,12 @@ function showhide(id) {
 }
 
 if (! $supported_file) {
-    die("<p>ERROR: file {$filename} does not exist</p>");
+    $error_message = "<p>ERROR: file {$filename} does not exist</p>";
+    if ($json) {
+        die(Json::invalidAPICall($error_message));
+    } else {
```
pascalchevrel: you could remove this else statement and simplify the code

pascalchevrel:
LGTM

views/translatestrings.inc.php
flodolo: We're loading the template if ´$json` is false, which means the resulting JSON will be malformed (didn't thought about it, realized only after trying). Also it might be a bit confusing returning JSON+400 when the request was originally for a web page.

PR #348, opened 2015-08-04
Add a tip about variable formats - fixes #311

### LINKED ISSUE #311: Add a tip about variable formats
We have a tip about {ok} tags in a yellow box on the locale page, Miroslav our Bulgarian localizer suggests that we also explain our replacement variables format because he found it confusing at first and thought that the ending s was a mistake.

It's probably a good idea as I have already seen others being confused with it and either removing the final s or translating the variable itself. Taking.

TheoChevalier:
Would this work? Still have to actually write the tip itself in listsitesforlocale.inc.php instead of the current placeholder.

Displaying the tip if:
- We caught a Python var error in the file (since we currently don't display any advice about what's wrong)
- There are Python variables in untranslated strings

Not displaying it if there is no variables in the file, or if the variables are all in translated strings without error.

### DIFF
--- a/classes/Langchecker/LangManager.php
+++ b/classes/Langchecker/LangManager.php
@@ -48,25 +48,24 @@ class LangManager
     {
         /* Test if all Python variables are present, analyzing both
          * format %(var)s or %s, or 'escaped' percentage sign (%%) */
-        $regex = '#%(\([a-z0-9._-]+\)s|[s%])#';
-        preg_match_all($regex, $reference, $matches_reference);
-        preg_match_all($regex, $translation, $matches_locale);
+        $matches_reference = Utils::getPythonVariables($reference);
+        $matches_locale = Utils::getPythonVariables($translation);
 
-        if ($matches_reference[0] != $matches_locale[0]) {
+        if ($matches_reference != $matches_locale) {
             /* Locale and reference have different variables. Count
              * instances of each variable and compare them. */
-            $count_occurences = function ($value, $array) {
-                // Count occurences of $value in $array
+            $count_occurrences = function ($value, $array) {
+                // Count occurrences of $value in $array
                 $count_values = array_count_values($array);
-                $occurences = isset($count_values[$value]) ?
+                $occurrences = isset($count_values[$value]) ?
                               $count_values[$value] :
                               0;
 
-                return $occurences;
+                return $occurrences;
             };
 
-            foreach ($matches_reference[0] as $python_var) {
-                if ($count_occurences($python_var, $matches_locale[0]) != $count_occurences($python_var, $matches_reference[0])) {
+            foreach ($matches_reference as $python_var) {
+                if ($count_occurrences($python_var, $matches_locale) != $count_occurrences($python_var, $matches_reference)) {
                     $errors[$reference] = [
                         'text' => $translation,
                         'var'  => $python_var,
@@ -75,8 +74,8 @@ class LangManager
             }
 
             // Check if locale has extra variables
-            foreach ($matches_locale[0] as $python_var) {
-                if (! in_array($python_var, $matches_reference[0])) {
+            foreach ($matches_locale as $python_var) {
+                if (! in_array($python_var, $matches_reference)) {
                     $errors[$reference] = [
                         'text' => $translation,
                         'var'  => $python_var,
--- a/classes/Langchecker/Utils.php
+++ b/classes/Langchecker/Utils.php
@@ -124,18 +124,32 @@ class Utils
     }
 
     /**
+     * Get Python variables from a string. Checking for
+     * %(var)s, %s, and %% ('escaped' percentage sign)
+     *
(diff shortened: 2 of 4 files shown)


classes/Langchecker/LangManager.php
```diff
@@ -48,15 +48,14 @@ public static function checkPythonErrors($reference, $translation, $errors)
...
-        preg_match_all($regex, $reference, $matches_reference);
-        preg_match_all($regex, $translation, $matches_locale);
+        $matches_reference = Utils::getPythonVar($reference);
+        $matches_locale = Utils::getPythonVar($translation);
 
-        if ($matches_reference[0] != $matches_locale[0]) {
+        if ($matches_reference != $matches_locale) {
             /* Locale and reference have different variables. Count
              * instances of each variable and compare them. */
             $count_occurences = function ($value, $array) {
```
flodolo: Can you fix also the typo in the variable name?

classes/Langchecker/Utils.php
```diff
@@ -124,18 +124,34 @@ public static function secureText($origin, $isarray = true)
...
      *
      * @param array $origin Original string
      *
-     * @return string String withon Python variables marked with <em>
+     * @return array Python variables
      */
-    public static function highlightPythonVar($origin)
+    public static function getPythonVar($origin)
     {
         $origin = htmlspecialchars($origin);
```
flodolo: I don't think you need this.

classes/Langchecker/Utils.php
```diff
@@ -124,18 +124,34 @@ public static function secureText($origin, $isarray = true)
...
+     *
+     * @param array $origin Original string
+     *
+     * @return string String with Python variables marked with <em>
+     */
+    public static function highlightPythonVar($origin)
+    {
+        $origin = htmlspecialchars($origin);
+        $matches = self::getPythonVar($origin);
+        foreach ($matches as $python_var) {
```
flodolo: No need to create `$matches` since the value is used only once

views/listsitesforlocale.inc.php
```diff
@@ -172,6 +172,14 @@
                                "     Example: </p><blockquote>;Plugins<br/>Plugins {ok}</blockquote>\n" .
                                "  </div>\n";
             }
+
+            if ($locale_analysis['untranslated_var'] || LangManager::countErrors($locale_analysis['errors'], 'python')) {
+                $todo_files .= "  <div class='tip'>\n" .
+                               "    <p><strong>Tip:</strong> This file contains variables that can use following formats:  \n" .
+                               "    </p><blockquote><strong>%(variable)s</strong>, <strong>%s</strong>, or <strong>%%</strong> (“escaped” percentage sign).</blockquote>\n" .
+                               "    <p>Make sure that they remain as they are in English.</p>\n" .
```
flodolo: Alternative text: `Make sure to keep these variables unchanged: don’t translate the text between brackets, or remove the “s” after the closing bracket.`

flodolo:
I like the idea of creating `Utils::getPythonVar`, we keep the regex and code only in one place. Also a lot cleaner referencing the variable and not the first element of the array. I would go for "getPythonVariables" though, and ignore naming consistency with the other function.

Having said that, I would display this warning only when there actually is a Python error, not when we have untranslated strings with variables. We display the tip about "{ok}" only when there are untranslated strings, to tell people they can "fix" the if they need it. 

So no need to add things to the data analysis structure, which is a piece we carry around everywhere and load thousands of times.

TheoChevalier:
Thanks for the review flod, I think I addressed all this ;)

views/listsitesforlocale.inc.php
```diff
@@ -172,6 +172,14 @@
                                "     Example: </p><blockquote>;Plugins<br/>Plugins {ok}</blockquote>\n" .
                                "  </div>\n";
             }
+
+            if (LangManager::countErrors($locale_analysis['errors'], 'python')) {
+                $todo_files .= "  <div class='tip'>\n" .
+                               "    <p><strong>Tip:</strong> This file contains variables that can use following formats:  \n" .
```
flodolo: One final nit: "use the following formats".

classes/Langchecker/Utils.php
```diff
@@ -124,18 +124,32 @@ public static function secureText($origin, $isarray = true)
...
+    /**
+     * Highlight Python variables in string
+     *
+     * @param array $origin Original string
+     *
+     * @return string String with Python variables marked with <em>
+     */
+    public static function highlightPythonVar($origin)
+    {
+        foreach (self::getPythonVariables($origin) as $python_var) {
```
flodolo: We need to keep `$origin = htmlspecialchars($origin);` here, otherwise the HTML code will be interpreted when displaying the wrong string.

On the other hand, we should be able to remove the call in getPythonVariables (I thought I commented on that one, but I'd need to check the old comments).

PR #354, opened 2015-08-18
Add script to remove active tag

flodolo:
I've been using it for a while, no point in leaving it out of the main repo.

### DIFF
--- a/scripts/mark_not_active
+++ b/scripts/mark_not_active
@@ -1,0 +1,48 @@
+#!/usr/bin/env php
+<?php
+namespace Langchecker;
+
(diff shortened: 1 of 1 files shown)


flodolo:
@TheoChevalier r?

TheoChevalier:
Tried one more just in case, and can't think of anything that's not needed. I took the opportunity to update the wiki https://github.com/mozilla-l10n/langchecker/wiki/CLI-scripts-syntax
With that, r+ from me :)

flodolo:
Thanks. This reminds me I need to add the export_surveys script

PR #367, opened 2015-09-10
Report extra strings in lang files as errors

flodolo:
Pootle and Pontoon sometimes try to store multiline strings in .lang files (localizers create new lines in the editor). Sometimes localizers forget to remove the English “translation” too.

While lang_update already removes them, it’s useful to have a check in the errors page.

### DIFF
--- a/classes/Langchecker/DotLangParser.php
+++ b/classes/Langchecker/DotLangParser.php
@@ -86,22 +86,29 @@ class DotLangParser
         // Make sure there's always an activation status to avoid controls later
         $dotlang_data['activated'] = false;
 
-        if ($file_content !== false) {
-            // First pass: extract strings and metadata (tags, active status) relevant for all locales.
-            $lines = count($file_content);
-            foreach ($file_content as $i => $current_line) {
-                if (! self::$extract_metadata && Utils::startsWith($current_line, '#')) {
-                    continue;
-                }
+        // Create an empty array for ignored strings to avoid controls later
+        $dotlang_data['errors']['ignoredstrings'] = [];
 
-                /* First line may contain an activation status
-                 * Tags are read with regexp "^## (\w+) ##", so trailing spaces can be ignored
-                 */
-                if ($i == 0 && rtrim($current_line) == '## active ##') {
-                    $dotlang_data['activated'] = true;
-                    continue;
-                }
+        if (! $file_content) {
+            return $dotlang_data;
+        }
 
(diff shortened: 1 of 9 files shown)


flodolo:
It took me a while to find a solution that doesn't make the errors view a lot slower (currently less than +100ms, all previous attempts where around +1500ms).

views/listsitesforlocale.inc.php
```diff
@@ -161,6 +161,16 @@
                     }
                     $todo_files .= "    </ul>\n";
                 }
+
+                if (LangManager::countErrors($locale_analysis['errors'], 'ignoredstrings')) {
+                    $todo_files .= "\n    <h3>{$current_filename}</h3><p>The following strings will be ignored:</p>\n";
```
pascalchevrel: We already have the filename displayed, I believe it should be:
```php
$todo_files .= "\n    <h3>The following strings will be ignored:</h3>\n";
```

classes/Langchecker/DotLangParser.php
```diff
@@ -86,10 +86,14 @@ public static function parseFile($path, $reference_locale = false)
...
         $dotlang_data['activated'] = false;
 
+        // Create an empty array for ignored strings to avoid controls later
+        $dotlang_data['errors']['ignoredstrings'] = [];
+
         if ($file_content !== false) {
-            // First pass: extract strings and metadata (tags, active status) relevant for all locales.
             $lines = count($file_content);
-            foreach ($file_content as $i => $current_line) {
+            for ($i = 0; $i < $lines; $i++) {
```
pascalchevrel: If you use a for loop, you might want to merge the previous line with it:
```php
for ($i = 0, $lines = count($file_content); $i < $lines; $i++) {
```

pascalchevrel:
It works well, my only worry is that this method is getting more and more long and complex over time, the only easy simplification I see would be to remove an indentation level  by doing that:
```php
    if (! $file_content) {
        return $dotlang_data;
    }

    for ($i = 0, $lines = count($file_content); $i < $lines; $i++) {
        ...blabla...
    }

    return $dotlang_data;
```
instead of:
```php
    if ($file_content !== false) {
        for ($i = 0, $lines = count($file_content); $i < $lines; $i++) {
            ...blabla...
        }
    }

    return $dotlang_data;
```

flodolo:
Fixed all comments, waiting for Travis.

I'll try to think a bit more if there's any way to simplify this method.

PR #373, opened 2015-09-17
Reorganize file structure to make switch to Twig easier

flodolo:
I think the ideal test would be to:
* Clone this in `~pascalc/langchecker_test` in the community server. 
* Copy over the existing `settings.inc.php`, add `$webroot_folder = '/~pascalc/langchecker/';` to it.
* Try it out.


flodolo:
I've been using this PR in my local VM for a few days, no issues so far. The main question remain to check if everything works as expected on the real server, given I've been bitten before for webstatus (and hopefully learned how to avoid that).

pascalchevrel:
It doesn't work with php dev server apparently

flodolo:
I confess that I never use the internal webserver, only Apache. 

It works for me with `php -S 0.0.0.0:8082` and leaving empty the `$webroot_folder`, which is a bit confusing. Will have to think a bit more about it.

pascalchevrel:
index.php should be in web folder

flodolo:
But at the point it will fail on the community server with Apache, I haven't managed to redirect with .htaccess to a file not in the root for webstatus.

pascalchevrel:
like https://l10n.mozilla-community.org/~pascalc/stores_l10n/ ?

pascalchevrel:
a quick look and I think that:
index.php should go to web/index.php (with the include path updated)
.htaccess  should go to web/.htaccess
line 44 in init.php, the  web.

pascalchevrel:
Just set up a virtualhost and it works with the changes above

flodolo:
Tried in a separate commit: dev server works, Apache doesn't unless the vhost is set on langchecker/web as root, and we don't have that on the server (or me locally).

pascalchevrel:
it will be a symlink to the web folder and the whole app will be outside of the public_html directory

pascalchevrel:
In functional tests, should now be:
define('INSTALL_ROOT',  realpath(__DIR__ . '/../../web/') . '/');

pascalchevrel:
merging and disabling momentarilly the cron  updating the server as it requires manual changes and I don't want it to be triggered while I am testing on the real server

Issue #379, opened 2015-09-24
Langchecker JSON should report errors like non UTF-8 encoding


flodolo:
If a file is completely localized, but it's not in UTF-8, the JSON doesn't report any error while it should.

PR #380, opened 2015-09-25
Expose UTF-8 errors in JSON

flodolo:
Fix issue #379.

Cost in terms of time is lower than expected (~20ms on my VM).


pascalchevrel:
That's because I improved the perf of isUTF8 in https://github.com/mozilla-l10n/langchecker/commit/b53ccdb8bfaa4bc6f834bef37c5cbc7739ac09c7

Issue #384, opened 2015-10-05
Changes needed for mozilla.org migration to GitHub


pascalchevrel:
- changes the source repositories
- updates links from viewvc to GitHub

PR #385, opened 2015-10-05
Issue #384: github migration update

pascalchevrel:
- change the source repository for mozilla.org
- updates links from viewvc to GitHub
- rewrite getSVNCommitTimestamp to getGitCommitTimestamp and hack it to ignore the renaming of the folder to en-US


classes/Langchecker/RawManager.php
```diff
@@ -34,7 +34,7 @@ public static function compareRawFiles($website, $locale, $filename)
             $reference_content = sha1_file($reference_filename);
             $results['reference_exists'] = true;
             $results['reference_url'] = Project::getPublicFilePath($website, $reference_locale, $filename);
-            $results['reference_lastupdate'] = Utils::getSVNCommitTimestamp($reference_filename);
+            $results['reference_lastupdate'] = Utils::getGitCommitTimestamp($reference_filename, $website[1]);
```
flodolo: Does it make sense to create a method in Project to retrieve `$website[1]` (like we have `Project::getWebsiteDataType`)? The idea is that this code shouldn't know how a website array is structured, and it makes our lives easier if we ever need to change that structure.

classes/Langchecker/Utils.php
```diff
@@ -181,21 +181,31 @@ public static function isUTF8($filename)
      *
      * @param string $filename File to analyze
      *
-     * @return int Timestamp of last commit from SVN,
+     * @return int Timestamp of last commit from git,
```
flodolo: Missing documentation for `$git_repo`.
Nit: Git.

flodolo:
Tested locally on my machine and can't find anything wrong.

Do you already have a date in mind for the switch?

PR #401, opened 2015-10-22
Move templates to Twig

flodolo:
This should be ready for review.

The only visual changes, besides heavy reorganization of the error page:
* Display deadline in the opt-in view.
* In listisitesforlocale: display the status (active, utf-8 error) also in todo files, display the utf-8 error more prominently (like missing strings).

Performances look similar to the existing ones.

PR #404, opened 2015-10-30
Optin redesign

flodolo:
Use checkboxes to select pages, use a button to open the URL

Reason for the redesign: nobody uses the link at the bottom of the page, I keep receiving multiple bugs filed at the same time from the same localizer.

I asked them and the link at the bottom is either unclear or not visible.

This is based on #401, so it will need to be rebased or fixed after that. Filing a PR to avoid forgetting about it.


flodolo:
Screenshots of how the page looks

schermata 2015-10-30 alle 19 32 15

PR #409, opened 2015-11-09
Reorganize sources, add script to update repos

flodolo:
One last update: make the repos storage structure a bit less cryptic, create a script that can be used to update all repos from a single cronjob.

I'm tempted to take it one step further, and store in settings.php only one or two paths (like GIT_PATH and SVN_PATH) like we do in Transvision, and eventually add a ´local_folder´ to the array to specify a different checkout folder.

Thoughts?

Based on #401.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -119,10 +119,23 @@ $repositories = [
 ];
 
 /*
- * If a file is not listed in $lang_flags, it's assumed to be non critical
- * for all locales.
+ * Flags are defined for each website later in the file. For each file in
+ * a website it's possible to specify flags, and for which locales these flags
+ * are valid.
+ * Currently we're using flags to tag files as critical, obsolete and opt-in.
+ *
+ * If a file is not listed, it's assumed to be non critical for all locales.
  * If a flag is valid for all locales, set it to ['all']. If it's not,
  * set the flag to the array of locales.
+ *
+ * Example (a file critical for French but opt-in for all other locales):
+ * $lang_flags['website_name'] = [
+ *     'file.lang' => [
+ *         'critical' => ['fr'],
+ *         'opt-in'   => ['all'],
+ *     ],
+ * ];
+ *
  */
 $lang_flags = [];
 
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -8,8 +8,12 @@ if (! isset($_SERVER['SERVER_NAME'])) {
 $settings_file = __DIR__ . '/settings.inc.php';
 if (! file_exists($settings_file)) {
     die('File app/config/settings.inc.php is missing. Please check your configuration.');
+} else {
+    require $settings_file;
+    if (! isset($local_storage)) {
+        die('$local_storage is missing in your configuration file. Please update app/config/settings.inc.php');
+    }
 }
-require $settings_file;
 
 // Real data is in adi.inc.php, not under VCS
 if (is_file(__DIR__ . '/adi.inc.php')) {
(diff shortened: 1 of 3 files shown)


flodolo:
To expand a bit further. `settings.inc.php` could look like this, with only one storage path
```
<?php
// Paths must end with slash. By default all repositories will be stored in this path
$local_storage = '/localpath/to/repositories/';

// Path to local clone of Locamotion's repo
$locamotion_repo  = '/localpath/to/git/mozilla-lang/';

/* This is is used to include assets in templates. Change this setting
 * according to your system: if langchecker is served from the root,
 * use '/', otherwise put the URL of the folder containing
 * langchecker, e.g. '/~pascalc/langchecker/'
 */
$webroot_folder = '/langchecker/';

const DEBUG = false;

/* It is possible to override local paths for specific repositories
 * by creating an array called $override_local with the following structure.
 *
 * $override_local = [
 *     'www.mozilla.org' => '/full/path/to/your/clone',
 * ];
 */

```

`source.inc.php` would look like this
```
if (! isset($override_local)) {
    // Make sure there is an array available to simplify checks
    $override_local = [];
}

$repositories = [
    'www.mozilla.org' => [
        'local_path'  => isset($override_local['www.mozilla.org']) ? $override_local['www.mozilla.org'] : $local_storage . 'git/www.mozilla.org',
        'public_path' => 'https://github.com/mozilla-l10n/www.mozilla.org/tree/master/',
        'repository'  => 'https://github.com/mozilla-l10n/www.mozilla.org',
        'vcs'         => 'git',
    ],

etc.
```

That way we're still free to have clones around in our local machines, but the prod server doesn't need all that flexibility.

pascalchevrel:
I think I will keep that for January if you don't mind, I am not much productive because of tirednesss these days since I came back from sick leave and I would like to focus on trying (hoping) to get my Q4 goals done and focus on features for our tools rather than refactoring them for the moment.

flodolo:
Sounds good to me. 

We still need to move the remaining SVN repositories to GitHub, and I believe doing that after this is fixed makes more sense. And the more I think about it, the more I like the idea of simplifying things per https://github.com/mozilla-l10n/langchecker/pull/409#issuecomment-164815699

flodolo:
I went ahead and added a second commit. This is how `settings.inc.php` would like on the community server to keep the current folder structure
```
$local_storage = ''/home/pascalc/';
$webroot_folder = '';

$override_local = [
    'www.mozilla.org'     => '/home/pascalc/repos/github/www.mozilla.org/',
    'start.mozilla.org'   => '/home/pascalc/startpagesvn/',
    'about:healthreport'  => '/home/pascalc/repos/github/fhr-l10n/',
    'engagement'          => '/home/pascalc/repos/github/engagement-l10n/',
    'add-ons'             => '/home/pascalc/repos/svn/l10n-misc/add-ons/',
    'firefox-updater'     => '/home/pascalc/repos/svn/l10n-misc/firefoxupdater/',
    'firefoxos-marketing' => '/home/pascalc/repos/svn/l10n-misc/firefoxos-marketing/',
    'appstores'           => '/home/pascalc/repos/github/appstores/',
];
```

flodolo:
I created a [copy of the langchecker installation on the server](https://l10n.mozilla-community.org/~pascalc/langchecker_repos/?action=errors), and everything seems to work as expected with the above config.

app/config/sources.inc.php
```diff
@@ -32,12 +23,107 @@
...
+    ],
+];
+
+/*
+ * If a file is not listed in $lang_flags, it's assumed to be non critical
+ * for all locales.
+ * If a flag is valid for all locales, set it to ['all']. If it's not,
+ * set the flag to the array of locales.
+ */
 $lang_flags = [];
```
TheoChevalier: Not exactly related to this patch, but maybe we should provide an example of which structure this array must follow? I'm also not sure what a flag really is, tbh :)

flodolo: There are flags for each website ([example for mozilla.org](https://github.com/flodolo/langchecker/blob/repos/app/config/sources.inc.php#L209)), this declaration is here only to avoid warnings later when setting up the website's specific flags. Right now we're only using 'critical', 'obsolete' (to hide them), and 'opt-in'.

flodolo:
Added a comment for the array, hope it's clear.

TheoChevalier:
Tested locally, works perfectly with my setup. Also tested with a local webdashboard pointing your demo server (even though testing the errors pages should be enough).
The demo server is really *slow*, but I assume this is only the lack of cache.

TheoChevalier:
(And of course, ship it)

flodolo:
Thanks, will merge tomorrow morning to make sure I have a caffeinated brain to avoid errors :coffee:

PR #420, opened 2015-12-28
untracked_files script: include reference, display relative path by default

flodolo:
Just noticed that firefox/independent.lang was removed from locales but not reference, which means it’s still exposed on Transvision and Pootle.

It’s better having the reference locale always included.

### DIFF
--- a/app/scripts/untracked_files
+++ b/app/scripts/untracked_files
@@ -3,19 +3,25 @@
 namespace Langchecker;
 
 $command_help = "untracked_files - List local lang files not tracked by langchecker\n" .
-                "Usage: untracked_files [websiteid] [locale]\n\n" .
+                "Usage: untracked_files [websiteid] [locale] [absolute]\n\n" .
+                "[websiteid] Website ID to analyze (e.g. 0 for mozilla.org)\n" .
+                "[locale]    Locale to analyze\n" .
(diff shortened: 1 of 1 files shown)


flodolo:
I'm actually going to do one more change, since now it's not possible to use the output to delete files like I used to do with SVN.

TheoChevalier:
I was looking for a way to delete files yesterday, checked the wiki, found nothing :) I think we should include this script in https://github.com/mozilla-l10n/langchecker/wiki/CLI-scripts-syntax 

Also, it’s not clear to me what I’m supposed to do once I have tens of absolute paths. I’m probably missing something, but wouldn’t it be easier if we were returning a full command we can copy/paste, like |rm /path/to/file && rm /path/to/file|, etc?

app/scripts/untracked_files
```diff
@@ -28,13 +34,12 @@ $current_website = $lang_based_sites[$cli_website];
 $reference_locale = false;
```
TheoChevalier: Looks like it can be removed

flodolo:
With SVN this script could be used as
```
~/langchecker/app/scripts/untracked_files | xargs svn del $1
```

With Git this doesn't work, so you move inside the repository's root and run
```
~/langchecker/app/scripts/untracked_files | xargs git rm $1
```

PR #426, opened 2016-01-12
Clean up obsolete files

flodolo:
Waiting for a couple of pull requests to be merged, leaving PR around in the meantime.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -44,7 +44,6 @@ $mozillaorg_lang = [
     'download.lang',
     'download_button.lang',
     'esr.lang',
-    'euballot.lang',
     'firefox/android/index.lang',
     'firefox/australis/firefox_tour.lang',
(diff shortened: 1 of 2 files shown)


flodolo:
@TheoChevalier Thoughts/comments on this massive cleanup?

TheoChevalier:
I think this makes a lot of sense. I assume the files will disappear from Pontoon/Pootle without any action needed? I must admit I had already forgotten some of those pages…

Is upgradepromos.lang still used?

flodolo:
Yes, they disappear automatically from both repos. I got complaints in the past from a few localizers about those files still be around.

Good point on `upgradepromos.lang`, I think we can also remove `upgradedialogs.lang` (it's stuff from fx36, based on the [SVN history](http://viewvc.svn.mozilla.org/vc/projects/mozilla.com/trunk/localesMOVEDTOGITHUB/en-US/upgradedialog.lang?view=log))

PR #436, opened 2016-02-03
Drop surveys as a standalone repository

flodolo:
Move only Hello survey to the engagement-l10n repository


flodolo:
Blocked by Pontoon's update.

PR #443, opened 2016-02-08
Reduce console output of lang_update and import_locamotion

flodolo:
Also update commenting style in some other scripts

### DIFF
--- a/app/classes/Langchecker/GetTextManager.php
+++ b/app/classes/Langchecker/GetTextManager.php
@@ -74,17 +74,16 @@ class GetTextManager
           'strings'  => [],
         ];
 
-        Utils::logger("== {$current_locale} ==");
-
         $local_import = $locamotion_repo != '' ? true : false;
 
-        /* Some locales need to me mapped because our systems and Pootle use
-         * different locale codes.
-         *
-         * Array format: local code => pootle code
-         *
-         * Use dashes and not underscores for Pootle, e.g. gn-PY (not gn_PY).
-         */
+        /*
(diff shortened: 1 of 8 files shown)


app/classes/Langchecker/GetTextManager.php
```diff
@@ -74,16 +74,15 @@ public static function importLocamotion($locale_data, $current_filename, $curren
...
-         *
-         * Use dashes and not underscores for Pootle, e.g. gn-PY (not gn_PY).
+        /*
+            Some locales need to me mapped because our systems and Pootle use
+            different locale codes.
+
+            Array format: Langchecker code => Pootle code
+
+            Use dashes and not underscores for Pootle, e.g. gn-PY instead of gn_PY.
          */
```
TheoChevalier: In https://github.com/mozfr/transvision/issues/609#issue-131635075 the ending comment section was aligned with the begining, not sure which one we should enforce everywhere (not critical, though :p) I wonder if PHP cs fixer could check that

flodolo: Uhm, I missed that. I confess that the misalignmed asterisks look a bit weird.

TheoChevalier:
otherwise, lgtm :beers:

app/classes/Langchecker/GetTextManager.php
flodolo: I see php.net has a [page with an example](http://php.net/manual/en/language.basic-syntax.comments.php) like Pascal's style, let's stick to that.

PR #444, opened 2016-02-09
Remove 'api' view

flodolo:


flodolo:
As far as I can tell, we're actively using [all views available](https://github.com/mozilla-l10n/langchecker/wiki/Available-URLs) in langchecker with the exception of this one, and `action=snippets` that I'm currently trying to push with the engagement team.

We already have a view to [get translation](https://github.com/mozilla-l10n/langchecker/wiki/Get-available-translations-in-JSON-format), and it's probably more intuitive than the `action=api` one.

@pascalchevrel 
This view already survived a couple of rewrites, but I'm not sure there's value in keeping it around. I also don't know much on the reasons why it was created, so I defer the decision to you. I'd be OK with closing this without merging.

flodolo:
@pascalchevrel Not sure if you saw this (more a decision than a code review).

PR #457, opened 2016-02-29
Fix App Store locales

TheoChevalier:


flodolo:
Makes sense, but I would update [this line](https://github.com/mozilla-l10n/langchecker/blob/master/app/config/sources.inc.php#L610) and [this one](https://github.com/mozilla-l10n/langchecker/blob/master/app/config/sources.inc.php#L220) to use `$ios_locales`. 

It also mean that we'd lose some locales on appstores, and we need to tell them (thinking of dsb, hsb).

TheoChevalier:
hsb and dsb do not get removed by ./untracked_files, and both repos are up-to-date, digging

PR #458, opened 2016-02-29
Update untracked script

flodolo:
List also folders for locales completely unknown.

### DIFF
--- a/app/scripts/untracked_files
+++ b/app/scripts/untracked_files
@@ -45,7 +45,7 @@ $ignored_files = [
     'privacy/notices/firefox-hello.lang',
     'privacy/notices/firefox.lang',
 ];
-$ignored_locales = ['ja'];
+$ignored_locales = ['hi', 'ja'];
 
 // Create list of locales to analyze
 if ($cli_locale == 'all') {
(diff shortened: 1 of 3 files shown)


flodolo:
@TheoChevalier r?

I still get one file on website 12 (Arabic)

app/scripts/untracked_files
```diff
@@ -45,7 +45,7 @@ $ignored_files = [
     'privacy/notices/firefox-hello.lang',
     'privacy/notices/firefox.lang',
 ];
-$ignored_locales = ['ja'];
+$ignored_locales = ['hi', 'ja'];
```
TheoChevalier: Why are we ignoring hi? There's probably a reason, but it's not obvious to me :)

flodolo: We don't support `hi`, we support `hi-IN`. But there's an activation file for legal-docs for the former.
https://github.com/mozilla-l10n/www.mozilla.org/tree/master/hi

TheoChevalier:
./untracked_files 0 returns a file for 'fur' locale, but we can't display untracked files for unsupported locales: ./untracked_files 0 fur returns Unsupported locale #fur.

Maybe we should change that?

TheoChevalier:
lgtm otherwise :snail:

flodolo:
Try updating www.mozilla.org? I [removed](https://github.com/mozilla-l10n/www.mozilla.org/commit/2f323617fa672291f18368066c43bba12046538a) the file about an hours ago, it's a remnant from PHP and it didn't make much sense to keep it

TheoChevalier:
My point was more about the different info returned regarding the different arguments, rather than the file itself :)

flodolo:
Well, I think it's by design: without specifying a locale we check the entire website, so you might get "unsupported" locales.

PR #464, opened 2016-03-12
Update App Store screenshots + whatsnew

TheoChevalier:
Do not merge yet
https://bugzilla.mozilla.org/show_bug.cgi?id=1255850
https://bugzilla.mozilla.org/show_bug.cgi?id=1252410

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -414,6 +414,7 @@ $appstores_lang = [
     'whatsnew/whatsnew_android_44.lang',
     'whatsnew/whatsnew_android_45.lang',
     'whatsnew/whatsnew_android_46_beta.lang',
+    'whatsnew/whatsnew_ios_2_1.lang',
 ];
 
 $lang_flags['appstores'] = [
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -414,6 +414,7 @@
     'whatsnew/whatsnew_android_44.lang',
     'whatsnew/whatsnew_android_45.lang',
     'whatsnew/whatsnew_android_46_beta.lang',
+    'whatsnew/whatsnew_ios_2_1.lang',
```
flodolo: Next version of iOS is 3.0, not sure where the 2.1 comes from.

TheoChevalier: It comes from the App Store, that's the version that has just been released

flodolo: Ah, I thought it was for the version we're localizing now. Also completely missed that the minor update for 2.0 was called 2.1

pascalchevrel:
I put the screenshots for v3.0 in a separate lang file so this invalidates this PR, I am going to close it.

PR #469, opened 2016-03-30
Add ESLint for JavaScript, improve scripts

flodolo:


flodolo:
@pascalchevrel @TheoChevalier 
Thoughts on the idea of using a linter for JavaScript? I kept separate commits to make things easier to follow. Also borrowed most of the ESLint configuration from bedrock.

Langchecker doesn't really have much JS, but it might be useful for larger projects like Transvision.

flodolo:
Tested introducing an error, and it [properly fails](https://travis-ci.org/mozilla-l10n/langchecker/jobs/119480461).

flodolo:
If you want to use ESLint locally, just install it with npm (`npm install -g eslint`). I'm currently using this [plugin](https://atom.io/packages/linter-eslint) for Atom.

TheoChevalier:
We're not heavy JS users, but I'd say that's nice to have, given that we can exclude external libs. :+1:

flodolo:
The command is currently running only on `web/assets/js` in Travis, and it's already excluding `sorttable.js`, so we can definitely ignore libraries.

PR #471, opened 2016-04-04
Remove Firefox OS marketing files from locale dashboards

TheoChevalier:
I noticed some files like screenshots_2_0.lang were still displayed on locale dashboards (for Arabic, for instance). This will help localizers set priorities and prevent them from translating stuff that won’t be used.
@flodolo does that look good to you?

I imagine we’ll remove the files at some point, but this is a first step


flodolo:
Yes, it makes absolutely sense.

I didn't merge them to the rest of engagement-l10n because I was planning to remove the entire repository at some point, but I guess we still need to figure out the story behind TV. In the meantime, definitely better to hide them.

PR #472, opened 2016-04-05
Remove en-US from $ios_locales

flodolo:
This caused https://github.com/mozilla-l10n/www.mozilla.org/commit/d7252554f194b68685ed82801a597181c4d2b933#diff-3057f915bb835cd221b0b1235ec14057

en-US shouldn't be in the list of locales supported for a page.

CC @TheoChevalier 
Not sure if you want to update the list also at https://github.com/mozilla-l10n/stores_l10n/blob/15633f598a78357575630fdc235f9cbccc4c6ed3/app/classes/Stores/Project.php#L43

I might fix WebStatus' API to not expose en-US for iOS at all.

### DIFF
--- a/app/config/locales.inc.php
+++ b/app/config/locales.inc.php
@@ -63,11 +63,12 @@ $firefox_desktop_android = array_merge($firefox_locales, $fennec_locales);
     https://github.com/mozilla-l10n/stores_l10n/blob/15633f598a78357575630fdc235f9cbccc4c6ed3/app/classes/Stores/Project.php#L43
 */
 $ios_locales = [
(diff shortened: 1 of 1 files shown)


TheoChevalier:
oops, thanks for fixing. I remember en-US listed with other locales was wanted on the appstore project, will ask pascal if it’s still relevant when he comes to the office today

flodolo:
Excluded locale from Webstatus API https://github.com/mozilla-l10n/webstatus/commit/4866d8b587687aeac0319950b0e3ed37de85e717

pascalchevrel:
en-US for Android should be kept as release engineering wants to publish English automatically from our API, for iOS, there is no automation so we can safely remove it.

PR #490, opened 2016-05-06
lang_update: when import is set, consider missing strings too

flodolo:


app/scripts/lang_update
```diff
@@ -104,16 +104,35 @@ foreach ($file_list as $current_filename) {
...
-            $common_strings = array_intersect(
-                array_keys($import_data['strings']),
-                array_keys($locale_data['strings'])
+
+        if (count($import_files) > 0) {
+            /*
+                Add to $locale_data strings that are available in the
+                reference file but are missing in the localized one. This
+                way we can import them before building the file.
+                Normally this is not necessary, since LangManager::buildLangFile
```
TheoChevalier: I assume "normally" is when we do not import anything?

flodolo: Yes. When you have to import anything, strings are added later when building the .lang file.

TheoChevalier:
Ok, LGTM :) I did a quick import test, this worked smoothly

flodolo:
Thanks.

PR #532, opened 2016-10-12
Sync up locales between web pages and FF for Android/iOS 

peiying2:


flodolo:
See the red error "All checks have failed"?
There's a Details link, taking you to
https://travis-ci.org/mozilla-l10n/langchecker/builds/167198935

Under Build jobs there are 2 red X, click on any of the two and scroll the page to identify the error
https://travis-ci.org/mozilla-l10n/langchecker/jobs/167198936

And I realized we already have 'kab' for the two landing pages, so we should remove it from the 2 shipping arrays.

There's one more change to do: check if any of the iOS locales you're adding is in `$ios_landing_page`, and then do the same for Android.

Example
https://github.com/mozilla-l10n/langchecker/blob/master/app/config/sources.inc.php
```
$ios_landing_page = array_unique(array_merge(
    $ios_locales,
    [
        'af', 'an', 'ar', 'ast', 'bn-BD', 'bn-IN', 'ca', 'el', 'en-GB',
        'es-AR', 'eu', 'fa', 'gl', 'gn', 'hi-IN', 'hu', 'ka', 'kab', 'kn',
        'lij', 'ml', 'ms', 'my', 'or', 'son', 'sq', 'sr',
    ]
));
```

This piece of code create an array of locales to use for the landing page by merging `$ios_locales` (the one you're changing) and another array of locales. There's no point in having duplicates, so you should remove ast, en-GB, hu, son. Leave 'kab' here instead of adding it to the `$ios_locales`.

Note that we need to coordinate to merge this.

flodolo:
Also need to remove duplicates from `$android_langing_page`: cak, es-CL, sr.

flodolo:
Final thought: this adds pages for a few locales, we should probably send a heads-up to dev-l10n-web that we're fixing things, and that's the result.

PR #538, opened 2016-10-20
Oct mo

peiying2:


peiying2:
Somehow my old commits are still showing up as part of this latest new pull request.  Only " 	remove 5 obsolete files" is not implemented.  My machine is up to date with the previous changes. It could have been I didn't create the pull requests for the older commits (syncing up the locales). Will try to commit for only the latest change (removal of old files).

flodolo:
I think the problem is that you used an obsolete branch. You can try this in langchecker
```
git checkout OctMo
git rebase master
git push -f origin OctMo
```
And then check how this PR looks.

flodolo:
Never mind, I see you opened a new PR. Anyhow, the problem is that master went ahead when you created the branch, so what you need is a "rebase against master". Since that rewrites history, you then need a forced push (`git push -f`).

PR #545, opened 2016-11-11
Add heartbeat/2016/nov2016.lang

peiying2:


flodolo:
This looks good 👍 

Ready to merge, as soon as we confirm the weird string.

peiying2:
Looks like the string appeared in the Sept version (they must have loved it):
Take your tabs on the road with Firefox for iOS and Android

TheoChevalier:
I assume “Your Firefox Account meets your phone. They fall in love.” is associated to “Make a match”? If yes, we should mention it because literal translations won’t work here

peiying2:
Yes, the button is next to the string.  should we move the two strings next to each other?

flodolo:
Reorganized the strings with new comments, still need to figure out what to do with the free range tabs 🐔

TheoChevalier:
> Yes, the button is next to the string. should we move the two strings next to each other?
We should add a note saying that this particular button is used with this specific string, and not with another one

PR #562, opened 2016-12-13
Add EOY mobile snippet

TheoChevalier:


app/config/sources.inc.php
```diff
@@ -376,9 +377,8 @@
     'snippets/2016/dec2016.lang' => [
         'critical' => ['de', 'es', 'fr', 'pt-BR', 'ru'],
     ],
-    'snippets/2016/dec2016_eoy_a.lang' => [ 'critical' => ['all'] ],
```
flodolo: Why did you remove these two? Marking as obsolete would be safer.

flodolo: Not really, if you run the script for untracked files you would end up removing them.

TheoChevalier: Hum, I’m not sure I follow you, I removed the flag from the files ($lang_flags['engagement']), not the tracking of the files ($engagement_lang). So, untracked_files returns nothing. Am I missing something else?

PR #578, opened 2017-01-18
Add setting to disable libmagic (fixes #312)

### LINKED ISSUE #312: Improve perf on errors page

flodolo:


flodolo:
Apparently there's no way to work around the error, even putting the finfo call into a try…except construct 👎 

At least this make it work without touching the code every time. Note that it's a performance hit on the errors page: on my system it goes from ~7 s to ~12 s.

TheoChevalier:
Pretty radical approach indeed, but works now and LGTM 👍

PR #579, opened 2017-01-19
Remove $disable_magic setting, properly catch exception in finfo()

flodolo:


flodolo:
This should be a saner approach:
* We can catch the exception, just need to deal with the Namespace (d'oh). 
* We need to silence warnings on the first call.

flodolo:
Sadly this doesn't work, errors page timeouts. The alternative would be to store if libmagic is functional once, and store that information, but given how rare is this case so far, better stick with the extra setting.

PR #580, opened 2017-01-19
Remove $disable_magic setting, properly catch exception in finfo()

flodolo:
Undo the last patch and catch exception on finfo(), store information about lib magic in a private static variable.


flodolo:
Let's try again…

flodolo:
I need to do one more pass on the logic to disable magic, I think I can simplify it.

flodolo:
Should be ready for testing/review.

PR #581, opened 2017-01-19
add new locales to newsletter; update deadline

peiying2:
- bug 1332132, add zh-HK and zh-TW to newsletter supported locales.
- bug 1331504, deadline to newsletter file


flodolo:
See tests failure. I completely missed that bug 1332132 was about zh-HK 👎 

You cannot add a locale that we don't support on mozilla.org for one file, that's not how it works, and it doesn't really make any sense to add a locale that will never be maintained (all pages will redirect to en-US). I'm afraid you need to find an alternative solution for this, like newsletter being available in zh-HK, but not the subscribe page.

flodolo:
Oops, didn't mean to close it.

app/config/sources.inc.php
```diff
@@ -768,8 +775,12 @@
         'mozorg/home/index-2016.lang'       => $mozillaorg,
         'mozorg/internet-health.lang'       =>
             [
-              'cy', 'de', 'es-ES', 'fr', 'kab', 'ko', 'pl', 'pt-BR',
+              'af', 'cy', 'de', 'es-ES', 'fr', 'kab', 'ko', 'pl', 'pt-BR',
             ],
+        'mozorg/internet-health.lang/privacy-security.lang'       =>
```
flodolo: Missing [

flodolo:
Issue in the current PR:
* Syntax is broken (=> not aligned)
* `mozorg/internet-health/privacy-security.lang` doesn't have the list of supported locales

app/config/sources.inc.php
```diff
@@ -771,8 +778,12 @@
...
```
flodolo: name is mispelled

PR #588, opened 2017-02-01
Add new locale to file; update deadline

peiying2:
1). add af to file per bug 1335320
2). change deadline per update in PR #4597


flodolo:
This looks good, but we need the PR adding strings first. You will also have to update the deadline for the home page if we add the new string there.

PR #589, opened 2017-02-01
Expose foundation/index

TheoChevalier:


TheoChevalier:
Correct, we’ll simulate a deadline for the vendor (just so that it’s close to a real request) and communicate it via Bugzilla

PR #591, opened 2017-02-07
Bugs 1336204 and 1337362

peiying2:
bug 1336204: March snippets
bug 1337362: add kab to two files


flodolo:
This looks good. Good job 👍 

We can merge as soon as the issues on engagement-l10n are fixed.

PR #596, opened 2017-03-02
Track Appstores files for 52 and 53 beta 

Delphine:
Not sure why the deadline for What's New skips from Firefox 51 to 52 in this file... doesn't make much sense to me. Updated with latest version and dates nevertheless.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -420,6 +420,8 @@ $appstores_lang = [
     'fx_android/whatsnew/android_51_beta.lang',
     'fx_android/whatsnew/android_51.lang',
     'fx_android/whatsnew/android_52_beta.lang',
+    'fx_android/whatsnew/android_52.lang',
+    'fx_android/whatsnew/android_53_beta.lang',
     'fx_ios/description_release.lang',
     'fx_ios/screenshots_v3.lang',
     'fx_ios/whatsnew/ios_2_1.lang',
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -446,13 +448,14 @@
     'fx_android/whatsnew/android_45.lang'      => [ 'obsolete' => ['all'] ],
     'fx_android/whatsnew/android_46.lang'      => [ 'obsolete' => ['all'] ],
     'fx_android/whatsnew/android_47.lang'      => [ 'obsolete' => ['all'] ],
-    'fx_android/whatsnew/android_48.lang'      => [ 'obsolete' => ['all'] ],
```
flodolo: Why are you removing 48 here? we should actually mark 49 and 50 here as obsolete (51 won't be for a few days more).

app/config/sources.inc.php
```diff
@@ -446,13 +448,14 @@
...
     'fx_android/whatsnew/android_46_beta.lang' => [ 'obsolete' => ['all'] ],
     'fx_android/whatsnew/android_47_beta.lang' => [ 'obsolete' => ['all'] ],
     'fx_android/whatsnew/android_48_beta.lang' => [ 'obsolete' => ['all'] ],
     'fx_android/whatsnew/android_49_beta.lang' => [ 'obsolete' => ['all'] ],
     'fx_android/whatsnew/android_50_beta.lang' => [ 'obsolete' => ['all'] ],
     'fx_android/whatsnew/android_51_beta.lang' => [ 'obsolete' => ['all'] ],
+    'fx_android/whatsnew/android_52_beta.lang' => [ 'obsolete' => ['all'] ],
+    'fx_android/whatsnew/android_53_beta.lang' => [ 'obsolete' => ['all'] ],
```
flodolo: 53_beta is definitely not obsolete

PR #597, opened 2017-03-02
Reorganize sources and support priority levels

flodolo:
Review can definitely wait next week, considering there are other PRs that need to add files (I'll integrate them as they're merged).

Would it help to have a PR updating the documentation too?


flodolo:
Looking into the test failure

flodolo:
OK, I was just lazy at updating the test config file.

flodolo:
Taking also the chance to cleanup some old files.
* download.lang
* esr.lang
* firefox/australis/fx36_tour.lang
* firefox/privacy_tour/privacy_tour.lang
* firefox/speed.lang

app/config/sources.inc.php
```diff
@@ -571,458 +49,1067 @@
...
-       5 reference locale,
-       6 url to public repo,
-       7 array of flags,
-       8 type of files (lang, raw)
-    ]
+    For each file it's possibile to specify the following fields:
+    * deadline: string, date in ISO format
+
+    * priorities: a multidimensional array, where each level of priority
+      (from 1 to 3) is associated to a list of locales. 'all' can be used to
```
TheoChevalier: Probably good to mention if 1 is the highest or the lowest priority (since Pontoon stars, I don’t want to make any assumptions regarding priorities :p)

flodolo:
I realize that we can drastically simplify the array for the most common situation, and use a similar approach for deadlines.

One priority for 'all' locales:
```
priority => 1,
```

Multiple priorities:
```
        'priorities'          => [
            1 => ['fr'],
            2 => ['all'],
        ],
```

We could even use 'priorities' for both cases, and check if the value is an array or integer, but the name might be confusing.

flodolo:
Per IRC discussion, using priority for both cases 👍 

Will start using separate commits from now on, in case you started looking at the code.

flodolo:
Updated code and doc.

app/config/websites.inc.php
```diff
@@ -0,0 +1,128 @@
...
+    12 => [
+        'appstores',
+        $repositories['appstores']['local_path'],
+        '',
+        // Added: ar, see https://bugzilla.mozilla.org/show_bug.cgi?id=1259200
+        array_unique(array_merge($google_play, ['ar'], $apple_store)),
+        $appstores_lang,
+        'en-US', // source locale
+        $repositories['appstores']['public_path'],
+        1,
```
TheoChevalier: Priority on Pontoon is 4 stars, do we want to match?

app/config/sources.inc.php
```diff
@@ -579,462 +49,1041 @@
...
+];
 
-    'about:healthreport' =>
-    [
-        'fhr.lang' => array_diff($firefox_desktop_android, $mozorg_locales),
+$firefoxhealthreport_lang = [
+    'fhr.lang' => [
+        'supported_locales' => array_diff($firefox_desktop_android, $mozorg_locales),
     ],
+];
```
TheoChevalier: Maybe worth indicating the default priority like for mozilla.org

app/config/sources.inc.php
```diff
@@ -579,462 +49,1041 @@
...
+        'supported_locales' => [
+            'de', 'es', 'fr', 'id', 'it', 'pl', 'pt-BR', 'ru',
+        ],
+    ],
+    'tiles/2016/tiles_jan2016.lang' => [
+        'supported_locales' => [
+            'de', 'es', 'fr', 'id', 'it', 'pl', 'pt-BR', 'ru',
+        ],
+    ],
+];
```
TheoChevalier: Maybe worth indicating the default priority like for mozilla.org

tests/testfiles/config/sources.php
```diff
@@ -1,16 +1,64 @@
...
-        'critical'  => ['fr'],
-        'obsolete'  => ['all'],
-        'testflag1' => ['all'],
-        'testflag2' => ['fr'],
+        'flags' => [
+            'obsolete'  => ['all'],
+            'testflag1' => ['all'],
+            'testflag2' => ['fr'],
+        ],
+        'priority' => [
```
TheoChevalier: I wonder if it could break if priority for `fr` is set first, and then overridden in the loop by `all`. Maybe it’s worth testing this?
`        'priority' => [`
`            3 => ['fr'],`
`            2 => ['all'],`
`        ],`
And this?
`        'priority' => [`
`            3 => ['it'],`
`            2 => ['all'],`
`            1 => ['fr'],`
`        ],`

tests/integrity/source_integrity.php
```diff
@@ -45,19 +45,83 @@
...
+                    $errors[] = "Deadline {$deadline} for {$filename} is of the wrong type (" . gettype($priority) . '). It should be a string.';
+                }
+            }
+
+            // Check priorities
+            if (isset($file_data['priority'])) {
+                $priority = $file_data['priority'];
+                if (gettype($priority) == 'integer') {
+                    // Single priority
+                    if ($priority < 1 || $priority > 3) {
```
TheoChevalier: `if ($priority < 1 || $priority > 5) {`

tests/integrity/source_integrity.php
```diff
@@ -45,19 +45,83 @@
...
+                        $errors[] = "Priority {$priority} for {$filename} is incorrect. It should be an integer between 1 and 3.";
```
TheoChevalier: 1 and 5

tests/integrity/source_integrity.php
```diff
@@ -45,19 +45,83 @@
...
+                    }
+                } elseif (is_array($priority)) {
+                    // Multiple priorities
+                    foreach ($priority as $key => $value) {
+                        if (gettype($key) != 'integer') {
+                            $errors[] = "Priority {$key} for {$filename} is of the wrong type (" . gettype($key) . '). It should be an integer between 1 and 3.';
```
TheoChevalier: 1 and 5

tests/integrity/source_integrity.php
```diff
@@ -45,19 +45,83 @@
...
+                        } else {
+                            if ($key < 1 || $key > 3) {
```
TheoChevalier: `if ($key < 1 || $key > 5) {`

tests/integrity/source_integrity.php
```diff
@@ -45,19 +45,83 @@
...
+                                $errors[] = "Priority {$key} for {$filename} is incorrect. It should be an integer between 1 and 3.";
```
TheoChevalier: 1 and 5

tests/integrity/source_integrity.php
```diff
@@ -45,19 +45,83 @@
...
+                            }
+                        }
+
+                        if (! is_array($value)) {
+                            $errors[] = "Priority {$key} for {$filename} should be assigned to an array of locales.";
+                        }
+                    }
+                } else {
+                    // Wrong type
+                    $errors[] = "Priority {$priority} for {$filename} is of the wrong type (" . gettype($priority) . '). It should be an integer between 1 and 3.';
```
TheoChevalier: 1 and 5

app/classes/Langchecker/Project.php
```diff
@@ -28,36 +28,26 @@ public static function getReferenceLocale($website)
     }
 
     /**
-     * Return supported files for website
+     * Return supported locales for a file in website. If not defined,
```
TheoChevalier: `for a file in $website`?

tests/units/Langchecker/Project.php
```diff
@@ -194,20 +202,41 @@ public function getWebsiteFilesDP()
...
+                        'deadline' => '2017-01-30',
+                        'priority' => 2,
+                    ],
+                    'file4.lang' => [
+                        'deadline' => [
+                            '2017-01-30' => ['fr'],
+                        ],
+                    ],
+                    'file5.lang' => [
+                        'deadline' => [
```
TheoChevalier: Same here as the logic in the function is basically the same as for priorities

app/config/sources.inc.php
```diff
@@ -579,462 +49,1041 @@
...
+
+      Different deadlines:
+
+      'deadline' => [
+          '2016-04-25' => ['de', 'fr'],
+          '2016-01-05' => ['all'],
+      ],
+
+    * priority: priorities are numeric values assigned to files. 1 is the
+      highest priority, 3 is the lowest.
```
flodolo: No, there are 3, like tiers. 

They're then mapped to a 5 stars system in Webdashboard for consistency with Pontoon (1 -> 5 stars, 2 -> 3 stars, 3 -> 1 star). Priority vs stars would be backwards anyway.

app/classes/Langchecker/Project.php
flodolo: Uhm, I actually meant to use "website" here, since we call them websites internally. Would project be less confusing?

app/config/sources.inc.php
flodolo: Interesting, I did add them, but apparently the last commit didn't make it to GH.

app/config/sources.inc.php
TheoChevalier: Understood that afterwards :/ Now I wonder why do we need to match tiers?

flodolo: We don't, it's just that I find Pontoon's classification extremely confusing. IMO if it's priority, 1 is more important than 5.

Since tiers use a similar system (1 -> 3), Pontoon will have to support mapping priority to stars anyway.

Alternatives are:
* Keep using this sytem
* Use priorities from 1 to 5, knowing that they're backwards compared to Pontoon
* Use the same system as Pontoon (5 -> 1)

app/classes/Langchecker/Project.php
TheoChevalier: I though you were referring explicitly to the first param, using website is fine

app/config/sources.inc.php
TheoChevalier: I personally don’t mind that it’s backwards, just find it confusing is the same project is 4 stars in Pontoon and 3 or 5 here :)

tests/testfiles/config/sources.php
flodolo: It can't break, because the code exits as soon as it finds a perfect match, while it keeps checking if it finds 'all' (that's actually what I'm testing). 

Having said that, having another file that tests that case wouldn't hurt.

app/config/sources.inc.php
flodolo: OK, let's use 5 them :-)

tests/testfiles/config/sources.php
TheoChevalier: Yup, seen that in the code, but it is fragile, I can see myself break the logic if I ever touch this part :p

flodolo: Not sure how you could break it. If you define 'fr' at some point, it will get that priority, independently from the order.

TheoChevalier: I meant if I edit the function, comments won’t prevent idiots like me to break the logic, but tests should :)

flodolo:
Added support for priorities from 1 to 5, added more tests on deadlines and priorities, also updated the webdashboard PR accordingly.

PR #598, opened 2017-03-02
add sv-SE to files per bug 1343949

peiying2:


flodolo:
I will add the files myself to avoid warnings on the dashboard.

PR #599, opened 2017-03-03
Emails + survey

TheoChevalier:


app/config/sources.inc.php
```diff
@@ -347,6 +350,7 @@
     'snippets/2017/mar2017.lang',
     'snippets/2017/mar2017_b.lang',
     'social/2016/fundraising.lang',
+    'surveys/copyright_stories.lang',
     'surveys/data_privacy_day.lang',
     'surveys/survey_hello_fx42.lang',
     'surveys/survey_maker_party_2016.lang',
```
flodolo: Do you want to get rid of this, per IRC discussion with Piotr?

TheoChevalier: Yeah, but will do a separate PR later, just want to get done with the current request

TheoChevalier: Later = after the refactor

PR #601, opened 2017-03-05
Remove old whatsnew pages for Android and iOS

flodolo:
Also stop using the hack for Arabic, they’re moving to beta next cycle

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -956,67 +956,7 @@ $appstores_lang = [
     ],
     'fx_android/description_release.lang' => [
         'deadline'          => '2016-04-30',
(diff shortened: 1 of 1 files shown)


flodolo:
@Delphine 
Any concern about doing this? Once merged I would remove also the files from the appstores repository.

flodolo:
Rationale for this: when we enable a new locale in Pontoon, they'll see all these files and probably start translating them, even if the webdashboard will hide them.

flodolo:
Thanks, merging and cleaning up.

I think we'll have to make sure ar, he, fa, ur are enabled on Pontoon for appstores after this.

PR #602, opened 2017-03-05
Remove engagement content for 2014-2015

flodolo:


flodolo:
@peiying2 
Any concern about doing this? Once merged I would remove also the files from the engagement-l10n repository.

I don't think there's a lot of value in keeping around content from 2 years ago, and it will make adding new files a bit easier.

peiying2:
Agreed.

PR #603, opened 2017-03-06
Remove Maker Party survey for Polish

TheoChevalier:
Not removing it for the other completed locales, because there are strings we can import in next surveys


flodolo:
I'll do the actual removal

TheoChevalier:
there’s a PR ;)

PR #604, opened 2017-03-07
Track What's New page for Focus 3.1

Delphine:


app/config/sources.inc.php
```diff
@@ -702,6 +702,13 @@
         ],
     ],
     'focus_ios/whatsnew/focus_2_1.lang' => [
+      'flags' => [
```
flodolo: Alignment is off for this part, but I'll fix it directly.

PR #605, opened 2017-03-07
add apr2017.lang in snippets

peiying2:
per bug 1345221

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -654,10 +654,10 @@ $engagement_lang = [
     'snippets/2017/mar2017_b.lang' => [
         'deadline'          => '2017-02-24',
         'supported_locales' => ['id', 'zh-TW'],
-
+    ],
     'snippets/2017/apr2017.lang' => [
         'deadline'          => '2017-03-21',
-        'supported_locales' => ['es-ES', 'de', 'fr', 'id', 'pt-BR', 'ru', 'zh-TW'],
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -654,6 +654,10 @@
     'snippets/2017/mar2017_b.lang' => [
         'deadline'          => '2017-02-24',
         'supported_locales' => ['id', 'zh-TW'],
+
+    'snippets/2017/apr2017.lang' => [
+        'deadline'          => '2017-03-21',
+        'supported_locales' => ['es-ES', 'de', 'fr', 'id', 'pt-BR', 'ru', 'zh-TW'],
```
flodolo: Remove the extra empty line before the code you're adding, and please fix the alphabetical order here ;-)

flodolo:
Oh, completely missed you ate a ], before.

It looks good now.

app/config/sources.inc.php
```diff
@@ -655,6 +655,10 @@
         'deadline'          => '2017-02-24',
         'supported_locales' => ['id', 'zh-TW'],
     ],
+    'snippets/2017/apr2017.lang' => [
+        'deadline'          => '2017-03-21',
+        'supported_locales' => ['de', 'es-ES', 'fr', 'id', 'pt-BR', 'ru', 'zh-TW'],
```
flodolo: I've just realized that we did 'es-ES' instead of 'es' here, not sure it made sense (the next month we went back to es).

PR #610, opened 2017-03-15
add snippet file apr2017_b.lang

peiying2:
per bug 1347295

### DIFF
--- a/app/config/locales.inc.php
+++ b/app/config/locales.inc.php
@@ -32,7 +32,7 @@ $fennec_locales = [
 
 // List of locales only working on mozilla.org
 $mozorg_locales = [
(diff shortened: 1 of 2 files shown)


app/config/locales.inc.php
```diff
@@ -32,7 +32,7 @@
 
 // List of locales only working on mozilla.org
 $mozorg_locales = [
-    'am', 'ncj', 'nv', 'zam',
```
flodolo: Were you planning to do this change in this PR?

flodolo: Given that the PR was already reviewed, I would have used a different PR.

At this point, keep it here, but note that you will have to a bunch of files on mozilla.org **and** engagement-l10n

PR #611, opened 2017-03-17
IH subpages: add new file; add locales to existing subpages; reorder files

peiying2:
bug 1348046 - kab
bug 1348065 - cy
bug 1344270 - new file: Open Innovation


flodolo:
Given that both kab and cy asked for the other two pages, I wonder if we should enable them for the new ones too (maybe ask in the bugs?)

PR #614, opened 2017-03-21
add firefox/all.lang; add new locales to IH subpages

peiying2:
Bug 1349127 - add af
Bug 1348780 - add pt-BR
Bug 1281144 - firefox/all.html

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -111,6 +111,10 @@ $mozillaorg_lang = [
         'deadline'          => '2016-03-15',
         'supported_locales' => $mozillaorg,
     ],
+    'firefox/all.lang' => [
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -111,6 +111,10 @@
         'deadline'          => '2016-03-15',
         'supported_locales' => $mozillaorg,
     ],
+    'firefox/all.lang' => [
+        'deadline'          => '2017-04-04',
+        'supported_locales' => $mozillaorg,
```
flodolo: We should probably use `$firefox_desktop_android` here, the same we use for firefox/new

We should also change priority: not sure if 1 or 2, definitely not 3 (default for mozilla.org)

PR #619, opened 2017-03-30
add new snippet entries per bug 1351798

peiying2:


TheoChevalier:
Note that you will need to enable the project on Pontoon for the new locales, and warn them about it

PR #622, opened 2017-04-04
Add Spanish + Polish Advocacy autoresponder

TheoChevalier:
No rush here either :)


TheoChevalier:
Oh wait… Reading the email once more, this is not the right email…

PR #627, opened 2017-04-15
Track AppStores files for 53 and 54 beta

Delphine:


flodolo:
Fixed indentation, also marked the previous release as obsolete since we're 3 days from release, and there's no point in completing it at this point
https://github.com/mozilla-l10n/langchecker/pull/627/commits/d61d4624cbdbc9f782671250cfbaf4ed45794694

PR #640, opened 2017-05-11
For PRs 4787, 4783, 4820 and 4798

peiying2:
- Two new IH subpages
- Update three existing IH subpages
- Update three Firefox subpages.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -103,7 +103,7 @@ $engagement_locales = [
 // Default priority is 3
 $mozillaorg_lang = [
     'download_button.lang' => [
-        'deadline'          => '2016-04-29',
+        'deadline'          => '2016-05-24',
         'priority'          => 1,
         'supported_locales' => $mozillaorg,
     ],
@@ -129,7 +129,7 @@ $mozillaorg_lang = [
         'supported_locales' => $firefox_locales,
     ],
     'firefox/channel/index.lang' => [
-        'deadline'          => '2017-04-25',
+        'deadline'          => '2017-05-24',
         'priority'          => 1,
         'supported_locales' => $mozillaorg, // Has Firefox for Android download buttons
     ],
@@ -437,8 +437,18 @@ $mozillaorg_lang = [
             'it', 'kab', 'ko', 'pl', 'pt-BR', 'sq', 'sv-SE', 'uk', 'zh-TW',
         ],
     ],
+    'mozorg/internet-health/decentralization.lang' => [
+        'deadline' => '2017-05-24',
+        'flags'    => [
+            'opt-in' => ['all'],
+        ],
+        'priority'          => 3,
+        'supported_locales' => [
+          'de', 'es-ES', 'fr',
+        ],
+    ],
     'mozorg/internet-health/digital-inclusion.lang' => [
-        'deadline' => '2017-04-10',
+        'deadline' => '2017-05-24',
         'flags'    => [
             'opt-in' => ['all'],
         ],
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -437,8 +439,18 @@
             'it', 'kab', 'ko', 'pl', 'pt-BR', 'sq', 'sv-SE', 'uk', 'zh-TW',
         ],
     ],
+    'mozorg/internet-health/decentralization.lang' => [
+        'deadline' => '2017-05-24',
+        'flags'    => [
+            'opt-in' => ['all'],
+        ],
+        'priority'          => 3,
```
flodolo: Why are you giving priority 3 to this while all the other have priority 1? Also note that priority 3 is the default for mozilla.org, you wouldn't need to set it explicitly.

app/config/sources.inc.php
```diff
@@ -437,8 +439,18 @@
...
+        'supported_locales' => [
+          'de', 'es-ES', 'fr',
```
flodolo: Last time you added cy and kab from the start. Do you want to do it this time too?

app/config/sources.inc.php
```diff
@@ -470,6 +482,16 @@
             'it', 'kab', 'ko', 'pt-BR', 'sq', 'sv-SE', 'uk', 'zh-TW',
         ],
     ],
+    'mozorg/internet-health/web-literacy.lang' => [
+        'deadline' => '2017-05-24',
+        'flags'    => [
+            'opt-in' => ['all'],
+        ],
+        'priority'          => 3,
```
flodolo: Same note here about priority and locales

app/config/sources.inc.php
```diff
@@ -103,7 +103,7 @@
 // Default priority is 3
 $mozillaorg_lang = [
     'download_button.lang' => [
-        'deadline'          => '2016-04-29',
+        'deadline'          => '2016-05-24',
```
flodolo: I would put a much shorter deadline for this, e.g. 2017-05-16. Note that we're in 📅 though ;-)

app/config/sources.inc.php
```diff
@@ -103,7 +103,7 @@
 // Default priority is 3
 $mozillaorg_lang = [
     'download_button.lang' => [
-        'deadline'          => '2016-04-29',
+        'deadline'          => '2016-05-17',
```
TheoChevalier: You probably want to update the year, too

app/config/sources.inc.php
```diff
@@ -129,8 +129,7 @@
         'supported_locales' => $firefox_locales,
     ],
     'firefox/channel/index.lang' => [
-        'deadline'          => '2017-04-25',
-        'priority'          => 1,
+        'priority'          => 2,
```
flodolo: Channel is definitely a priority 1 file

app/config/sources.inc.php
```diff
@@ -219,9 +218,11 @@
         'supported_locales' => $firefox_desktop_android,
     ],
     'firefox/nightly_firstrun.lang' => [
-        'flags' => [
+        'deadline'          => '2017-05-24',
+        'flags'             => [
             'opt-in' => $firefox_locales,
         ],
+        'priority'          => 1,
```
flodolo: I don't think the Nightly firstrun page should be priority 1, maybe 2?

app/config/sources.inc.php
flodolo: If you want this file to be 3, you can remove this line, it will get 3 by default

app/config/sources.inc.php
flodolo: Same here, remove this line

PR #643, opened 2017-05-12
add locale pbb; add pt-BR to foundation files

peiying2:
bug 1364397 (pt-BR)
add pbb to basic mozorg files per email request.


peiying2:
I left line 151 as is. The locale codes are out of order towards the end.

PR #646, opened 2017-05-17
Updated footer for Polish

TheoChevalier:
A bit overkill, but we might add more locales, so this will be handy (Also Piotr asked for it)


TheoChevalier:
Thanks!
Those are the strings on the footer of emails, and they are not translated in Polish yet :) We will need this translated if we support additional locales for newsletters. Not much additional context

PR #648, opened 2017-05-22
Bug 1363041 - Add support for Fennec nightly

flodolo:
Also clean up obsolete deadlines


flodolo:
It's expected for tests to fail. They search for a "nightly" API from stores_l10n that won't be available until https://github.com/mozilla-l10n/stores_l10n/pull/126 is merged and deployed

PR #649, opened 2017-05-22
Add opt-in pages to Pontoon locales

flodolo:
Also set cak as desktop locale

CC @peiying2


flodolo:
Other issues found:
* an wasn't set as Pootle locales
* pbb had a bunch of untracked files

PR #651, opened 2017-05-23
add locales dsb, hsb, sw to files

peiying2:
Bugs 1367180, 1367179, and 1367162

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -398,7 +398,7 @@ $mozillaorg_lang = [
             'opt-in' => ['all'],
         ],
         'priority'          => 2,
-        'supported_locales' => $mozilla,
+        'supported_locales' => $mozillaorg,
     ],
     'mozorg/contribute/index.lang' => [
         'flags' => [
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -19,7 +19,7 @@ $getinvolved_locales = [
     'el', 'en-GB', 'es-AR', 'es-CL', 'es-ES', 'es-MX', 'fa', 'fr', 'fy-NL',
     'he', 'hi-IN', 'hr', 'hsb', 'id', 'it', 'kab', 'ko', 'lt', 'ms', 'ncj',
     'nl', 'nv', 'pl', 'pt-BR', 'pt-PT', 'ro', 'ru', 'sl', 'son', 'sq', 'sr',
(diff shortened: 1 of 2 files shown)


app/config/sources.inc.php
```diff
@@ -350,7 +350,7 @@
             'en-ZA', 'es-AR', 'es-CL', 'es-ES', 'es-MX', 'eu', 'fa', 'fr',
             'fy-NL', 'hi-IN', 'hsb', 'it', 'ja', 'kab', 'km', 'ko', 'lt', 'ms',
             'ncj', 'nl', 'nv', 'pl', 'pt-BR', 'pt-PT', 'ro', 'ru', 'son', 'sq',
-            'sr', 'sv-SE', 'tr', 'uk', 'ur', 'zam', 'zh-CN', 'zh-TW',
+            'sr', 'sv-SE', 'sw', 'tr', 'uk', 'ur', 'zam', 'zh-CN', 'zh-TW',
```
flodolo: Let's not add it here, it's an extremely old page.

app/config/sources.inc.php
```diff
@@ -404,7 +404,7 @@
             'es-MX', 'eu', 'fa', 'fi', 'fr', 'fy-NL', 'gd', 'gl', 'hi-IN', 'hr',
             'hsb', 'hu', 'id', 'it', 'ja', 'kab', 'km', 'ko', 'lt', 'mk', 'ms',
             'ncj', 'nl', 'nv', 'pl', 'pt-BR', 'pt-PT', 'ro', 'ru', 'sk', 'sl',
-            'son', 'sq', 'sr', 'sv-SE', 'tr', 'uk', 'uz', 'vi', 'xh', 'zam',
+            'son', 'sq', 'sr', 'sv-SE', 'sw', 'tr', 'uk', 'uz', 'vi', 'xh', 'zam',
```
flodolo: I'm starting to think that we should open this to all mozilla.org locales.

app/config/sources.inc.php
peiying2: should i remove some other recently added locales, am, ncj, nv

app/config/sources.inc.php
peiying2: changed to 
'supported_locales' => $mozilla,

app/config/sources.inc.php
flodolo: No, at this point I would leave them be.

peiying2:
@flodolo sw has been in the supported locale list a while back. Is that what you are referring to?

PR #654, opened 2017-06-01
add file mozorg/about/governance/policies/participation.lang

peiying2:
Per bug 1364268.

NOTE: 
- I listed es-ES as the locale, since the file lives in the corresponding repo. The file is technically in "es".

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -363,6 +363,11 @@ $mozillaorg_lang = [
         'priority'          => 2,
         'supported_locales' => $mozillaorg,
     ],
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -363,6 +363,12 @@
         'priority'          => 2,
         'supported_locales' => $mozillaorg,
     ],
+    'mozorg/about/governance/policies/participation.lang' => [
+        'priority'          => 3,
```
flodolo: No need to add priority 3, it's the default for mozilla.org files
https://github.com/mozilla-l10n/langchecker/blob/master/app/config/sources.inc.php#L103

PR #658, opened 2017-06-05
add locale pai to mozilla.org project per bug 1369804

peiying2:


flodolo:
Note: after you add the files, you can move the entire folder to production with meld, no need to use cherry-pick if you want.

peiying2:
Thanks Flod.

PR #660, opened 2017-06-09
Add new Internet Health pages

flodolo:


flodolo:
mozorg/internet-health/index.lang has the same locales enabled for all other pages.

mozorg/internet-health/shared.lang has all locales enabled for internet-health/privacy-security.lang, since it's the one with most locales

mozorg/internet-health/shared.lang is also not supposed to be activated, since it's a shared file.

Turns out this page is obsoleted by mozorg/internet-health/index.lang
https://github.com/mozilla/bedrock/pull/4859/files#diff-76a34c96123a80eeeaf7c511d01086bb

If the new file is activated, we display the new hub. Removing the opt-in tag for now, we'll need to obsolete it soon (take a note @peying2?).

flodolo:
I remove the opt-in flag on shared, to avoid locales asking only for that, since it's short…

We need to keep in mind to enable it, if not already enabled, if there's a request to add any of the new internet-health pages.

PR #661, opened 2017-06-09
add locale sl to 9 files per bug 1371791

peiying2:
Also included the IH shared file.


flodolo:
Looks good, I will add the files myself to trunk.

PR #662, opened 2017-06-12
Add new locale azz; add sv-SE to IH pages

peiying2:
Bug 1371902 - Add sv-SE to IH pages
Bug 	1370966 - Enable azz


flodolo:
Let's remove ncj too

flodolo:
I've already moved the folder and added the missing files for sv-SE, just run gitup to update your local VM.

PR #666, opened 2017-06-16
Fix cell alignment on total line

TheoChevalier:


flodolo:
Not sure how we missed it for so long…

PR #672, opened 2017-07-06
update deadline per PR4938

peiying2:
- update nightly message in firefox/channel/index.lang


flodolo:
I think we should stop adding deadlines for this kind of changes.

Reason: it's not a real deadline (unlike the new pages we pushed out), we publish the string when it's translated.

At the same time, these deadlines help us figure out when the page was last updated…

Thoughts?

peiying2:
@flodolo I thought of the same thing when I added the line. Without it, one would not know which one was recently updated.  However, given the high priority rating, hopefully this helps guide localizers what to do first if they don't check dashboard often.

PR #673, opened 2017-07-07
Add script to enable shared features/products pages

flodolo:


flodolo:
@peiying2 
This script does the right thing: it activates the new pages only if firefox/shared.lang is completely localized. Simply run mozup and it will take care of activating pages where it's possible.

For this reason I've updated the script, removing mark_active for all these pages, make sure to not add them back
https://github.com/mozilla-l10n/vm_scripts/commit/e938c127520ec52226d396a84dc8fef5a8491aba

If we'll have more of these pages in the future, depending on `firefox/shared.lang`, I'll need to update this script.

PR #674, opened 2017-07-17
add snippets aug2017 per bug 1381592

peiying2:
Want to get this out, independent from the mozilla.org stuff. This is more straight forward.


app/config/sources.inc.php
```diff
@@ -894,6 +894,12 @@
         'deadline'          => '2017-06-19',
         'supported_locales' => ['de'],
     ],
+    'snippets/2017/aug2017.lang' => [
```
flodolo: I have the feeling this name choice is going to bite you in a couple of weeks, since this sounds like an out of cycle request 🐕

PR #676, opened 2017-07-19
Add developer.lang to the list

peiying2:
Add firefox/products/developer.lang to the activated file list per PR172.


flodolo:
There's no point in doing this alone, you need to do this when you add the file to langchecker (it's not tracked at all yet) ;-)

PR #677, opened 2017-07-21
Update file per mozilla.org PR #172

peiying2:
- add onboarding.lang
- add firefox/products/developer.lang
- update deadline to onboarding, developer and sync files to 8/20
- change file order under firefox/products/...

NOTE:
- to be obsolete developer.lang supports firefox locales, not mozillaorg.
- new firefox/products/developer.lang is set to be supported by mozillaorg locales. should we change?
- did not change priority for sync or new developer file.  should we? Sync 1 and dev lower?


flodolo:
You forgot to add the developer file to the new script for firefox_shared

PR #682, opened 2017-07-26
add Pocket email per bug 1384394

peiying2:
Pocket email is for Polish only.


flodolo:
Merging since you merged the other PR…

PR #684, opened 2017-07-28
Track Appstores files for 55 and beta56

Delphine:


app/config/sources.inc.php
```diff
@@ -1045,10 +1048,19 @@
...
         'supported_locales' => $fx_android_store,
     ],
     'fx_android/whatsnew/android_55_beta.lang' => [
+        'flags' => [
+            'obsolete' => ['all'],
+        ],
+        'supported_locales' => $fx_android_store,
+    ],
+    'fx_android/whatsnew/android_55.lang' => [
+        'deadline'          => '2017-08-02',
```
flodolo: Deadline for this content doesn't change: it's still Aug 7 (when we release, not when we merge)

PR #691, opened 2017-08-03
Per mozorg PR 175

peiying2:
add two new files
obsolete one  
update sendto locale list

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -322,7 +322,7 @@ $mozillaorg_lang = [
     ],
     'firefox/sendto.lang' => [
         'priority'          => 1,
-        'supported_locales' => $mozillaorg,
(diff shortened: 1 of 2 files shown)


app/config/sources.inc.php
```diff
@@ -317,7 +322,7 @@
     ],
     'firefox/sendto.lang' => [
         'priority'          => 1,
-        'supported_locales' => $firefox_locales,
+        'supported_locales' => $mozillaorg,
```
flodolo: This is **NOT** part of this PR (never been). Leave the set of language as it is.

peiying2: 活蛇添足 :-\

flodolo: For the record: Google Translate to Italian -> "add enough live snake"…

PR #695, opened 2017-08-11
Show errors on tags for en-US

flodolo:


flodolo:
@peiying2 
The errors page now shows errors like the one from this morning. Looks like we had one lingering for a while :-\
https://github.com/mozilla-l10n/www.mozilla.org/commit/19e0a307eff58680dbba8e485cbaa12e42bb075d

PR #705, opened 2017-09-04
mozilla.org updates

flodolo:
* Added firefox/new/sem.lang variant (German only)
* Added firefox/whatsnew_56.lang (all locales)
* Marked firefox/sync.lang as obsolete, replaced by firefox/features/sync.lang
* Update deadlines for several files


app/config/sources.inc.php
```diff
@@ -298,7 +302,11 @@
...
     ],
     'firefox/whatsnew_50.lang' => [
-        'deadline'          => '2017-05-08',
+        'priority'          => 1,
+        'supported_locales' => $firefox_locales,
+    ],
+    'firefox/whatsnew_56.lang' => [
+        'deadline'          => '2017-09-26',
         'priority'          => 1,
         'supported_locales' => $firefox_locales,
```
flodolo: Is this supposed to be German only?

flodolo: Confirmed it's for all languages.

PR #715, opened 2017-09-22
add 2 new files, add locale eo to file

peiying2:
PR5118: quantum.lang for de
PR5121: developer-quantumn.lang
bug 1402209: add eo to firefox/nightly_whatsnew.lang

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -247,7 +247,7 @@ $mozillaorg_lang = [
             'opt-in' => $firefox_locales,
         ],
         'supported_locales' => [
-            'bn-BD', 'bs', 'cak', 'cs', 'cy', 'de', 'dsb', 'en-GB', 'es-CL',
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -263,6 +263,10 @@
         'supported_locales' => $mozillaorg,
     ],
     'firefox/products/developer.lang' => [
+        'deadline'          => '2017-10-06',
```
flodolo: Why are you setting a deadline for this page? You should remove this and mark it as obsolete instead.

app/config/sources.inc.php
```diff
@@ -276,6 +280,11 @@
         'priority'          => 1,
         'supported_locales' => $mozillaorg,
     ],
+    'firefox/quantum.lang' => [
+        'deadline'          => '2017-09-26',
+        'priority'          => 1,
+        'supported_locales' => ['de'],
```
flodolo: Question: should this be open to other languages at some point?

peiying2: I asked and I was told explicitly this is for de only.  There will be a "global" quantum copy ready on Oct 3.

PR #722, opened 2017-10-10
add new file FxnU_oct.lang per bug1407074

peiying2:
Resort the order of the files.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -744,12 +744,16 @@ $engagement_lang = [
         'deadline'          => '2017-01-27',
         'supported_locales' => ['de', 'fr'],
     ],
+    'emails/2017/focus_android.lang' => [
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -744,12 +744,16 @@
...
+        'deadline'          => '2017-06-19',
+        'supported_locales' => ['pl'],
+    ],
     'emails/2017/fx_mobile_update.lang' => [
         'deadline'          => '2017-09-22',
         'supported_locales' => ['pl'],
     ],
-    'emails/2017/focus_android.lang' => [
-        'deadline'          => '2017-06-19',
+    'emails/2017/fxnU_oct.lang' => [
```
flodolo: Please don't spare characters, and don't use uppercase in file names. I guess 'firefoxandyou_oct.lang' would be good.

PR #728, opened 2017-10-19
add /firefox/hub/home-quantum.lang

peiying2:
Also remove en-US/firefox/hub/home.lang


flodolo:
Why are you adding this in a PR, when the file itself is not tracked? Please use one PR for all the work on langchecker ;-)

I'm going to close this one for now.

PR #731, opened 2017-10-24
add new file nov2017_mobile per bug 1410997

peiying2:


app/config/sources.inc.php
```diff
@@ -979,6 +979,12 @@
             'cs', 'de', 'es', 'fr', 'id', 'pt-BR', 'ru', 'zh-TW',
         ],
     ],
+    'snippets/2017/nov2017_mobile.lang' => [
+        'deadline'          => '2017-11-10',
+        'supported_locales' => [
+            'cs', 'de', 'es', 'fr', 'id', 'it', 'pt-BR', 'ru', 'zh-TW',
```
flodolo: 'cs' is not requested in the bug, 'pl' is and it's missing.

app/config/sources.inc.php
```diff
@@ -976,7 +976,13 @@
     'snippets/2017/oct2017.lang' => [
         'deadline'          => '2017-10-25',
         'supported_locales' => [
-            'cs', 'de', 'es', 'fr', 'id', 'pt-BR', 'ru', 'zh-TW',
+            'de', 'es', 'fr', 'id', 'pl', 'pt-BR', 'ru', 'zh-TW',
```
flodolo: You updated the wrong file

PR #751, opened 2017-12-12
add nn-NO to file per bug 1422650

peiying2:
* mozorg/internet-health/index.lang


flodolo:
Does it make any sense to translate the landing page without all the subpages?

PR #760, opened 2018-01-05
PR216 - add new file, update deadlines

peiying2:


app/config/sources.inc.php
```diff
@@ -392,6 +392,10 @@
         'priority'          => 1,
         'supported_locales' => $mozillaorg,
     ],
+    'mozorg/500.lang' => [
+        'priority'          => 1,
```
flodolo: I would leave this as priority 3 (remove the line), it's not a page they should see…

PR #776, opened 2018-03-02
Email review mode

TheoChevalier:
Goal: Provide something [similar to this](https://labs.cynthiapereira.com/l10n/email-donor_update_feb.html) to localizers for reviewing emails. So that they don’t have to generate it manually. It should help catch typos before reviewing the test send.

Some notes:
- No need to review this today (even, shouldn’t :) )

- I haven’t added extra logic to exclude other .lang files, because the view still looks good for other files, even if it’s designed with emails in mind.
For the sake of testing, I’m adding tags to the next email (emails/2018/en-US/donor_survey_march.lang), which will likely get exposed tonight.

- The 45 first lines of reviewstrings.php are duplicated from translatestrings.php. Before moving code around, I wanted to check with you on where you would prefer to store this code. New shared controller file?

- I’m using `|raw` in the template to get something close to a real email, I hope there’s no security concerns.

- One side effect is that tags are displayed next to strings in the translate view, that’s a positive side effect, at least when it comes to copy/pasting emails. I might even add extra tags that are not used in the review view, for the sake of identifying some strings more easily.

Let me know what you think about this approach when you have a moment! Hope it’s not too awful

### DIFF
--- a/app/config/websites.inc.php
+++ b/app/config/websites.inc.php
@@ -65,8 +65,9 @@ $repositories = [
        5 reference locale,
        6 url to public repo,
        7 default priority,
-       8 type of files (lang, raw)
-       9 project name on Pontoon
+       8 type of files (lang, raw),
+       9 project name on Pontoon,
+       10 array of excluded error checks
     ]
 */
 
@@ -83,6 +84,7 @@ $sites =
         3,
         'lang',
         'mozillaorg',
+        [],
     ],
     6 => [
         'engagement',
@@ -95,6 +97,7 @@ $sites =
         1,
         'lang',
         'engagement',
+        ['tags'],
     ],
     12 => [
         'appstores',
@@ -107,5 +110,6 @@ $sites =
         2,
         'lang',
         'appstores',
+        ['tags'],
     ],
 ];
--- a/app/config/websites.inc.php
+++ b/app/config/websites.inc.php
@@ -67,7 +67,7 @@ $repositories = [
        7 default priority,
        8 type of files (lang, raw),
        9 project name on Pontoon,
-       10 array of excluded error checks
+       10 array of excluded error checks (tags)
     ]
 */
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: false
 language: php
 php:
-  - 5.6
   - 7.0
+  - 7.2
 before_install:
   ## Disable xdebug
   - phpenv config-rm xdebug.ini
--- a/app/classes/Langchecker/Project.php
+++ b/app/classes/Langchecker/Project.php
@@ -454,6 +454,9 @@ class Project
                 case 'translate':
                     $result['file'] = 'translatestrings';
                     break;
+                case 'review':
+                    $result['file'] = 'reviewstrings';
+                    break;
             }
 
             return $result;
--- a/app/classes/Langchecker/Project.php
+++ b/app/classes/Langchecker/Project.php
@@ -87,6 +87,19 @@ class Project
     }
 
     /**
+     * Check if a type of errors is excluded from a website
+     *
+     * @param array  $website  Website data
+     * @param string $type Error type
+     *
+     * @return boolean True if the error check is excluded from this website
+     */
+    public static function isExcludedErrorCheck($website, $type)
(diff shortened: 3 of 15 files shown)


flodolo:
Without looking in depth at the code, one big concern is about errors:
* Need to declare tags at the beginning of the en-US like a standard mozilla.org page
* Need to add them to complete localized files. But right now `add_tags` only works on active files, and both `add_tags` and `mark_active` only work on website 0

At that point it would probably make sense to exclude a file or website from errors, ideally without hard coding the information, i.e. add it to the configuration in `$sites`.

flodolo:
> At that point it would probably make sense to exclude a file or website from errors, ideally without hard coding the information, i.e. add it to the configuration in $sites.

To clarify: only exclude from checks related to tags.

flodolo:
I did a quick test of the patch before looking at the code, and I think it makes sense. Before going in review mode, I'd suggest to:
* Allow a website to be excluded from errors related to tags. This requires adding a parameter, methods, and update docs. Maybe something like `errorchecks_exclusions`, and have an array of possible values (`tags` in this case). I don't think it would be a lot of code.
* Move the shared code up to line 45 into a separate file.

> I’m using |raw in the template to get something close to a real email, I hope there’s no security concerns.

I don't think it's a problem, we control the content displayed (it comes from our repos anyway).

> One side effect is that tags are displayed next to strings in the translate view, that’s a positive side effect, at least when it comes to copy/pasting emails.

You lost me. That's been there for a while, both as a list at the beginning, and next to each string bound to a tag ([example](https://l10n.mozilla-community.org/langchecker/?website=0&file=firefox/new/quantum.lang&action=translate)).

TheoChevalier:
> I did a quick test of the patch before looking at the code, and I think it makes sense. Before going in review mode, I'd suggest to:
> 
>     Allow a website to be excluded from errors related to tags. This requires adding a parameter, methods, and update docs. Maybe something like errorchecks_exclusions, and have an array of possible values (tags in this case). I don't think it would be a lot of code.

Thanks! Will look into this. Totally forgot to check error management, tbh

>     Move the shared code up to line 45 into a separate file.

Sounds good, mostly wanted to make sure this was what you wanted too

> 
>     One side effect is that tags are displayed next to strings in the translate view, that’s a positive side effect, at least when it comes to copy/pasting emails.
> 
> You lost me. That's been there for a while, both as a list at the beginning, and next to each string bound to a tag (example).

Sorry, not saying that’s new, only that it’s a nice perk that comes along with using tags in emails, for staff extracting translations into Salesforce

TheoChevalier:
Of course it wouldn’t have been so simple, I did in fact change something in the first 45 lines, and forgot about it… I’m grouping translations by locale, not by string ID. Added a Boolean before including the shared file.

TheoChevalier:
Also not sure where you wanted my to name the array `errorchecks_exclusions`, but if it was in websites.inc.php, it would have made things more complicated(?)

flodolo:
> Also not sure where you wanted my to name the array errorchecks_exclusions, but if it was in websites.inc.php, it would have made things more complicated(?)

Probably not super clear, but `errorchecks_exclusions` was an example of name for a method (which is dumb, since we use camelCase for those).

Anyhow, I had in mind exactly what you did: an additional parameter to the website, with methods to retrieve exclusions.

I'll review the code tomorrow before EOD, from a quick pass it looks great 👍

TheoChevalier:
Updated the Wiki here https://github.com/mozilla-l10n/langchecker/wiki/Available-URLs/476a257ad4549ba36003dc1f915496f04c7c6b07
Updating the l10n-drivers documentation repo is next

app/config/websites.inc.php
```diff
@@ -65,8 +65,9 @@
        5 reference locale,
        6 url to public repo,
        7 default priority,
-       8 type of files (lang, raw)
-       9 project name on Pontoon
+       8 type of files (lang, raw),
+       9 project name on Pontoon,
+       10 array of excluded error checks
```
flodolo: Add `(tags)` at the end, since it's the only supported exclusion at the moment

TheoChevalier:
Thanks, should be fixed!

PR #805, opened 2018-05-07
add May email file per bug1458771

peiying2:


app/config/sources.inc.php
```diff
@@ -730,6 +730,10 @@
         'deadline'          => '2018-02-18',
         'supported_locales' => ['de', 'es', 'fr', 'pl', 'pt-BR'],
     ],
+    'emails/2018/firefox_learning_journey.lang' => [
+        'deadline'          => '2018-06-03',
```
flodolo: Actually, this doesn't match the date in the bug (Jun 1st). Do they really want a deadline on Friday? Monday would be definitely better…

peiying2: Yes, I purposely add the weekend in the schedule. The "production" of the localized copy starts on Monday. I will comment in the bug.

PR #820, opened 2018-06-12
Track AppStores files for 61 and 62 beta

Delphine:


app/config/sources.inc.php
```diff
@@ -1134,13 +1134,19 @@
         'supported_locales' => $fx_android_store,
     ],
     'fx_android/whatsnew/android_59.lang' => [
+        'flags' => [
+            'obsolete' => ['all'],
+    ],
```
flodolo: nit: This `]` should be indented, starting on same row as `'flags'`

PR #822, opened 2018-06-20
Newsletter per bug 1467252

peiying2:
- German only per email team
- Other locales (pt-br, ru, id, pl, de, es, fr) could be supported at a much later time. 
- This file is different from the other newsletter file which has a different set of locales.


app/config/sources.inc.php
```diff
@@ -496,6 +496,11 @@
         'priority'          => 1,
         'supported_locales' => ['hi-IN'],
     ],
+    'mozorg/newsletter/opt-out-confirmation.lang' => [
```
flodolo: Is this the name of the file in the PR?
https://github.com/mozilla-l10n/www.mozilla.org/pull/234/files

PR #825, opened 2018-06-29
add new file, update deadlines per PR235

peiying2:
#5755: MOSS update, hi-IN only
#5791: New, Privacy FAQ page; update Index page
#5793: Add form field strings to newsletter page
#5850 (replaces 5777): Privacy landing page update (no deadline)
#5857: Privacy update on Firefox channel pages (no deadline)

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -487,7 +487,7 @@ $mozillaorg_lang = [
         'supported_locales' => $mozillaorg,
     ],
     'mozorg/moss/index.lang' => [
-        'deadline'          => '2017-07-30',
+        'deadline'          => '2018-07-15',
         'priority'          => 1,
         'supported_locales' => ['hi-IN'],
     ],
@@ -497,7 +497,7 @@ $mozillaorg_lang = [
         'supported_locales' => ['hi-IN'],
     ],
     'mozorg/newsletters.lang' => [
-        'deadline'          => '2017-03-27',
+        'deadline'          => '2018-07-15',
         'priority'          => 2,
         'supported_locales' => $newsletter_locales,
     ],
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -522,6 +522,11 @@
         'priority'          => 1,
         'supported_locales' => $mozillaorg,
     ],
+    'privacy/faq.lang' => [
+        'priority'          => 2,
+        'deadline'          => '2018-07-30',
+        'supported_locales' => $mozillaorg,
```
flodolo: I don't think this is correct. The privacy main page (privacy/index.lang) is for a much smaller set of locales.

peiying2: There is a page called privacy/principles.lang which supports all mozorg locales, so I thought this should be at the same level.

flodolo: Have you tried opening it? It redirects me to https://www-dev.allizom.org/internet-health/privacy-security/, which makes me think it was removed at some point.

flodolo: > Have you tried opening it? It redirects me to https://www-dev.allizom.org/internet-health/privacy-security/, which makes me think it was removed at some point.

Never mind, that's not the principles page.

peiying2: You mean this one: https://www.mozilla.org/it/privacy/principles/

flodolo: It's not discoverable without the index (I've updated the other comment in the meantime).

Maybe it makes sense to open the other page to all locales, with the caveat of not being able to translate policies. Anyhow, your call ;-)

peiying2: I hate to make decisions at this time of the day!! 
How discoverable is the "principles". I have a feeling the FAQ page could be linked from other places down the road. Can we make it an opt-in page?

peiying2: I will leave this one as is, and decide on the Index page after talking to legal. It doesn't look good for sure.

peiying2: @flodolo ok to proceed otherwise?

flodolo:
I'm going to add the files to the mozilla.org PR and merge. Given how old it is, there's a high risk of conflicts, and it's late for me too.

PR #828, opened 2018-07-05
add locale ia to file per bug 1473267

peiying2:


flodolo:
Worth asking if he wants more pages, to avoid doing this every 2 days?

PR #831, opened 2018-07-17
change privacy index page locale list

peiying2:
@flodolo Legal gave an ok to open up side panel to all locales. However, the Mozilla PN (main page) will be limited to the currently supported languages. I left the deadline the same as FAQ page.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -529,7 +529,8 @@ $mozillaorg_lang = [
     ],
     'privacy/index.lang' => [
         'priority'          => 1,
-        'supported_locales' => array_diff($legal_locales, ['et']),
+        'deadline'          => '2018-07-30',
+        'supported_locales' => $mozillaorg,
     ],
     'privacy/principles.lang' => [
         'priority'          => 2,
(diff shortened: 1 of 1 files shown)


flodolo:
It's quite a different thing to say the can and they should localize it. Should it be opt-in, instead of open to all locales? I don't have a strong opinion, since it's not a large file

peiying2:
Possible options given legal changed priority languages:
- Only for priority locales (all though Mozilla PN is localized in more than the priority languages)
- Opt-in 
- Open to all (this will save our time with individual request for a small file)

What do you think?

flodolo:
> What do you think?

It's your call ;-)

peiying2:
@flodolo pmac confirmed that activating this page won't mess up anything, :).

flodolo:
> @flodolo pmac confirmed that activating this page won't mess up anything, :).

I'm sure about that. The question is if we're going to show that page for, let's say 'nn-NO', when there are no legal docs. Does the file need to be activated for that? Will it happen anyway?

flodolo:
For sake of unblocking this, I've tested it myself: you need to remove the file from `$no_active_tag`. Once activated, other locales seem to load the page correctly.

PR #852, opened 2018-11-30
add new file per PR6528 (CPG update)

peiying2:


app/config/sources.inc.php
```diff
@@ -373,6 +373,11 @@
             'de', 'es-ES', 'fr', 'hi-IN', 'ja', 'pt-BR', 'zh-TW',
         ],
     ],
+    ' en-US/mozorg/about/governance/policies/reporting.lang' => [
```
flodolo: en-US?

app/config/sources.inc.php
```diff
@@ -373,7 +373,7 @@
             'de', 'es-ES', 'fr', 'hi-IN', 'ja', 'pt-BR', 'zh-TW',
         ],
     ],
-    ' en-US/mozorg/about/governance/policies/reporting.lang' => [
+    ' mozorg/about/governance/policies/reporting.lang' => [
```
flodolo: What about the extra leading whitespace? :-\

PR #878, opened 2019-04-29
Add mozorg/whats-is-a-browser.lang

alexgibson:


peiying2:
It looks good. Do you want to add deadline or priority for the focused languages? Both are optional.

alexgibson:
I think this is fine as-is. The page is just SEO content, so no hard deadline.

PR #879, opened 2019-04-30
remove locales bn-BD, bn-IN

peiying2:
I saw these two are still in the mozilla.org locale list and thought they should be removed. 

Also, I have not been able to see mozilla.org/bn go live.


flodolo:
Interesting, I'm sure I had a commit to fix this, but apparently it got stuck on my machine.

> Also, I have not been able to see mozilla.org/bn go live.

What does it mean? It's live, they just have too many pages untranslated
https://www.mozilla.org/bn/firefox/accounts/
https://www.mozilla.org/bn/privacy/

peiying2:
I will write to the community to help them prioritize.

PR #881, opened 2019-05-14
Add /firefox/new/trailhead.lang

alexgibson:


peiying2:
Looks good. I will need to send an email regarding the deadline for non-critical locales. If these locales finish it by June 4th, will they see their pages on production? Or they will be ready by Fx 68 release.

alexgibson:
> If these locales finish it by June 4th, will they see their pages on production?

If all goes to plan they should see their pages in production by June 4 yes.

PR #890, opened 2019-06-26
Add /firefox/best-browser.lang

amychurchwell:


peiying2:
@amychurchwell the request looks fine. Is there a deadline or priority level you want to add? Also @craigcook is working on another new file. Let's coordinate so we can release both at the same time. I

amychurchwell:
@peiying2 There is no deadline or priority for this one so I left those fields blank. 🙂

PR #901, opened 2019-07-18
Clean up the list of locales for engagement

flodolo:
Open question for nb-NO:
- There is one file, but it's not translated
- It was never enabled in Pontoon

Should it be kept and added to Pontoon instead?

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -45,7 +45,7 @@ $android_landing_page = array_unique(array_merge(
 
 $engagement_locales = [
     'bg', 'cs', 'cy', 'da', 'de', 'en-GB', 'es-ES', 'es', 'fr', 'hi-IN', 'id',
-    'it', 'nl', 'pl', 'pt-BR', 'ru', 'sl', 'sv-SE', 'sw', 'zh-TW',
+    'it', 'ja', 'nl', 'pl', 'pt-BR', 'ru', 'sl', 'sv-SE', 'sw', 'zh-TW',
 ];
 
 /*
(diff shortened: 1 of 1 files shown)


TheoChevalier:
I’m confused, did `ja` got disabled from Pontoon yesterday by mistake, or it wasn’t enabled in the first place? At the same time, there is translated content in the `ja` folder…

flodolo:
> I’m confused, did `ja` got disabled from Pontoon yesterday by mistake, or it wasn’t enabled in the first place? At the same time, there is translated content in the `ja` folder…

Yes, Japanese was removed from Pontoon by mistake, I'll add it back. Stage still has all the locales.
https://mozilla-pontoon-staging.herokuapp.com/projects/engagement/

nb-NO was enabled, but they never translated MoFo email. Should we keep it?

TheoChevalier:
> > I’m confused, did `ja` got disabled from Pontoon yesterday by mistake, or it wasn’t enabled in the first place? At the same time, there is translated content in the `ja` folder…
> 
> Yes, Japanese was removed from Pontoon by mistake, I'll add it back. Stage still has all the locales.
> https://mozilla-pontoon-staging.herokuapp.com/projects/engagement/

Ok, thanks!

> 
> nb-NO was enabled, but they never translated MoFo email. Should we keep it?

No, totally fine to drop

flodolo:
> nb-NO was enabled, but they never translated MoFo email. Should we keep it?

Never mind, saw your reply on Slack about this one. Japanese should be restored in the next sync on Pontoon.

PR #904, opened 2019-07-24
Add whatsnew_69

craigcook:
Lang file added in https://github.com/mozilla-l10n/www.mozilla.org/pull/326

Targeting "Tier One" locales (English, French, German) for Fx69 but we'll soon have a new evergreen default whatsnew page for all locales.


craigcook:
D'oh! I was mistaken. The intent is to send this to all locales after all. I'll have another PR shortly to expand the list.

PR #916, opened 2019-09-13
add lockwise page

Errolyn:
l10n lang file: https://github.com/mozilla-l10n/www.mozilla.org/pull/329


peiying2:
If this page can be opened up to more locales, you can create an extra entry for opt-in.

flodolo:
Please don't merge changes to langchecker before the page is available on the server.

PR #919, opened 2019-09-19
Add whatsnew 70

amychurchwell:
Ref: https://github.com/mozilla-l10n/www.mozilla.org/pull/332

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -317,7 +317,7 @@ $mozillaorg_lang = [
         'supported_locales' => $firefox_locales,
     ],
     'firefox/whatsnew_70.lang' => [
-        'deadline'          => '2019-10-14',
+        'deadline'          => '2019-10-11',
         'priority'          => 1,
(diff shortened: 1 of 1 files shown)


amychurchwell:
@peiying2 Mozilla staff are going to localize DE and FR, do I need to declare that in langchecker or just coordinate with you?

peiying2:
> @peiying2 Mozilla staff are going to localize DE and FR, do I need to declare that in langchecker or just coordinate with you?

We have communicated to the respective communities on this and more pages before October launch. There is no way to do that from langchecker, :).

peiying2:
The PR looks good. Could you confirm if the date is the due date to Releng for a full list of locales to be included?

amychurchwell:
@peiying2 
I updated the due date based on the schedule I was given of:
- **Go live - Friday October 11th**
- QA (external) - Mon Oct 14
- Firefox 70 Rapid Release - Tues Oct 22

I was told that all other locales other than DE and FR were opt-in.

PR #927, opened 2019-10-01
Fixing locales for lockwise

Errolyn:
Fixing the locales so they are inline with the intended coverage.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -229,7 +229,9 @@ $mozillaorg_lang = [
     'firefox/products/lockwise.lang' => [
         'deadline'          => '2019-10-20',
         'priority'          => 1,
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -227,12 +227,9 @@
         'supported_locales' => $firefox_locales,
     ],
     'firefox/products/lockwise.lang' => [
-        'flags' => [
-            'opt-in' => $mozilla,
-        ],
         'deadline'          => '2019-10-20',
         'priority'          => 1,
-        'supported_locales' => $key_market_locales,
+        'supported_locales' => ['de', 'en-CA', 'en-GB', 'es-ES', 'fr', 'it',]
```
peiying2: Please follow the same format for `supported locales` for other files.

PR #928, opened 2019-10-02
Update navigation.lang, add footer.lang

craigcook:
Strings added in https://github.com/mozilla-l10n/www.mozilla.org/pull/334

The new nav strings appear in a separate Firefox navbar that will be English-only until translation is complete.

The footer strings are all new and will be English-only until translation is complete.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -3,6 +3,7 @@
 $no_active_tag = [
     'download_button.lang',
     'firefox/shared.lang',
+    'footer.lang',
     'main.lang',
     'mozorg/internet-health/shared.lang',
     'navigation.lang',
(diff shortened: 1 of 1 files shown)


peiying2:
The behavior of a shared file is that there is no activation involved. The Nov 18 date must be set on Bedrock side. Or we can check a few weeks later to see how many locales are completed (number of locales or user %) and decide if we want to switch to the new design sooner than specified date.

craigcook:
> The behavior of a shared file is that there is no activation involved. The Nov 18 date must be set on Bedrock side. Or we can check a few weeks later to see how many locales are completed (number of locales or user %) and decide if we want to switch to the new design sooner than specified date.

OK, I wasn't sure... I saw the date on navigation.lang so I updated it to six weeks from now and used the same date for footer, but I don't think that actually has much meaning. We'll just have to watch the status and update bedrock when we think we have good enough coverage.

PR #931, opened 2019-10-09
Add firefox/welcome/page2.lang

craigcook:
Strings in https://github.com/mozilla-l10n/www.mozilla.org/pull/340

This will be for French and German, translated in-house.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -278,12 +278,10 @@ $mozillaorg_lang = [
             3 => ['all'],
         ],
         'deadline'          => '2019-11-25',
-        'supported_locales' => array_diff(
-            $key_market_locales,
-            [
-                'it', 'ja', 'ko', 'nl', 'pl', 'pt-BR', 'pt-PT', 'ru',
-                'zh-CN', 'zh-TW',
-            ]),
+        'supported_locales' => [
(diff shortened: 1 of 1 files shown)


peiying2:
@craigcook From Eric's email response on locale support (need clarity thread), he mentioned that for this page, 15 locales should be enabled. Could you check with him if this is the case?

craigcook:
OK, I talked to Eric. The Pocket app is available in 15 languages, so we would only want to show this page to those. But as with other marketing pages we'll do French and German in-house and community can opt in to other locales with a longer deadline. I'll update this PR.

app/config/sources.inc.php
```diff
@@ -272,9 +272,17 @@
...
-        'supported_locales' => $key_market_locales,
+        'priority'          => [
+            1 => ['de', 'fr', 'en-GB', 'en-CA'],
+            3 => ['all'],
+        ],
+        'deadline'          => '2019-11-25',
+        'supported_locales' => array_diff(
+            $key_market_locales,
+            [
+                'it', 'ja','ko', 'nl', 'pl', 'pt-BR', 'pt-PT', 'ru',
```
peiying2: need an extra space between ja and ko

craigcook: Oops!

peiying2:
Looks good. Please proceed with the PR.

PR #932, opened 2019-10-09
Add firefox/home-master.lang

craigcook:
Initially just for French and German for in-house translation. We expect there may be changes to this page in the near future and we can expand to more locales once things settle down.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -175,7 +175,7 @@ $mozillaorg_lang = [
         'supported_locales' => $mozillaorg,
     ],
     'firefox/home-master.lang' => [
-        'deadline'          => '2019-10-20',
+        'deadline'          => '2019-11-25',
         'priority'          => 1,
         'supported_locales' => $mozillaorg,
     ],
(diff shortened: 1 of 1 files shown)


peiying2:
The Key markets include es-ES too which we send out to agency to complete.

craigcook:
I've been told there are NOT any immediate plans to change this page after skyline (no plans, that is... it'll probably end up with unplanned changes, but such is life). So we can expose this to all locales, though we'll still handle French and German in-house.

Should we extend the deadline?

peiying2:
If that's the case, let's change the deadline.

craigcook:
Pushed deadline to 2019-11-25 (about six weeks)

flodolo:
We should stop adding the deadline in this file. The only consumer of that information was the webdashboard, which has been decommissioned months ago, and the deadline needs to be manually added in Pontoon these days. Same goes for priority.

craigcook:
> We should stop adding the deadline in this file. The only consumer of that information was the webdashboard, which has been decommissioned months ago, and the deadline needs to be manually added in Pontoon these days. Same goes for priority.

Ah, I had no idea. Noted, and I'll let the others know that this info isn't needed (at least not in this file).

peiying2:
True. It's more for my record of what to set in Pontoon. Maybe we can remove the lines of code but include in the comment in the PR, so I have the info for Pontoon.

PR #941, opened 2019-11-08
mobile-2019

Errolyn:
The demo is here: https://bedrock-demo-errolyn.oregon-b.moz.works/en-US/firefox/mobile/
The bedrock PR is here: mozilla/bedrock#7957
The lang file PR: https://github.com/mozilla-l10n/www.mozilla.org/pull/346


peiying2:
Do you have a proposed deadline so I can put in Pontoon?

Errolyn:
Looks like the deadline is 11/25, also I just found out that DE and FR are coming from staff, do you need me to change anything with that info?

PR #942, opened 2019-11-14
Add welcome pages 3 and 4

craigcook:
Prioritizing "tier 1" locales. We'll be getting French and German translations in-house. All others are opt-in.

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -278,7 +278,7 @@ $mozillaorg_lang = [
     ],
     'firefox/welcome/page2.lang' => [
         'priority'          => [
-            1 => ['de', 'fr', 'en-GB', 'en-CA'],
+            1 => ['de', 'fr', 'en-GB', 'en-CA',],
             3 => ['all'],
         ],
         'deadline'          => '2019-11-25',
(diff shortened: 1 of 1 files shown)


peiying2:
@craigcook Is welcome/3 due on Nov 20? And welcome/4 on Dec 11? Want to double confirm.

craigcook:
> @craigcook Is welcome/3 due on Nov 20? And welcome/4 on Dec 11? Want to double confirm.

Correct. We're getting FR and DE provided in-house for both, and those are the target launch dates for these pages in those locales. Other locales can opt in at their leisure.

app/config/sources.inc.php
```diff
@@ -287,6 +287,20 @@
             'pt-BR', 'pt-PT', 'ru', 'zh-CN', 'zh-TW',
         ],
     ],
+    'firefox/welcome/page3.lang' => [
+        'priority'          => [
+            1 => ['de', 'fr', 'en-GB', 'en-CA'],
```
peiying2: needs a `,` after the last locale entry.

app/config/sources.inc.php
```diff
@@ -287,6 +287,20 @@
...
+        ],
+        'deadline'          => '2019-11-20',
+        'supported_locales' => $mozillaorg,
+    ],
+    'firefox/welcome/page4.lang' => [
+        'priority'          => [
+            1 => ['de', 'fr', 'en-GB', 'en-CA'],
```
peiying2: needs a `,` after the last locale entry.

peiying2:
@craigcook sorry, my bad, what you did the first time was correct. The closing bracket is on the same line of the last locale, the comma is not needed. Please revert back to the prior version, then I will merge.

craigcook:
Oh looks like that trailing comma is only needed when there's a line break

PR #951, opened 2019-12-17
Add firefox/compare.lang

craigcook:


app/config/sources.inc.php
```diff
@@ -112,6 +112,13 @@
         'priority'          => 1,
         'supported_locales' => $firefox_desktop_android,
     ],
+    'firefox/compare.lang' => [
```
peiying2: put this entry after firefox/channel/index.lang

PR #953, opened 2020-01-16
Add community-hotline.lang

craigcook:
Lang file added in https://github.com/mozilla-l10n/www.mozilla.org/pull/362

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -419,13 +419,13 @@ $mozillaorg_lang = [
         'priority'          => 3,
         'supported_locales' => $mozillaorg,
     ],
+    'mozorg/about/governance/policies/community-hotline.lang' => [
+        'supported_locales' => $participation_locales,
+    ],
     'mozorg/about/governance/policies/participation.lang' => [
         'supported_locales' => $participation_locales,
     ],
     'mozorg/about/governance/policies/reporting.lang' => [
-        'supported_locales' => $participation_locales,
-    ],
-    'mozorg/about/governance/policies/community-hotline.lang' => [
         'supported_locales' => $participation_locales,
     ],
     'mozorg/about/history-details.lang' => [
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -424,6 +424,11 @@ $mozillaorg_lang = [
             'de', 'es-ES', 'fr', 'hi-IN', 'ja', 'pt-BR', 'zh-TW',
         ],
     ],
+    'mozorg/about/governance/policies/community-hotline.lang' => [
+        'supported_locales' => [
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -416,6 +416,11 @@
             'de', 'es-ES', 'fr', 'hi-IN', 'ja', 'pt-BR', 'zh-TW',
         ],
     ],
+    'mozorg/about/governance/policies/community-hotline.lang' => [
+        'supported_locales' => [
+            'de', 'es-ES', 'fr', 'hi-IN', 'ja', 'pt-BR', 'zh-TW',
```
peiying2: per bug https://bugzilla.mozilla.org/show_bug.cgi?id=1600148#c8 - the locale list should expand to 

 ar, de, es-ES, fr,  hi, id, it, ja, nl, pl, pt-BR, ru,  zh-CN, zh-TW

This means all three CPG files need to be updated with the locale list

craigcook: Ah, I missed that. Thanks!

peiying2:
Let's change `hi` to `hi-IN`. On mozilla.org, we only have hi-IN.

app/config/sources.inc.php
```diff
@@ -408,12 +408,17 @@
     ],
     'mozorg/about/governance/policies/participation.lang' => [
         'supported_locales' => [
-            'de', 'es-ES', 'fr', 'hi-IN', 'ja', 'pt-BR', 'zh-TW',
+            'ar', 'de', 'es-ES', 'fr', 'hi', 'id', 'it', 'ja', 'nl', 'pl', 'pt-BR', 'ru', 'zh-CN', 'zh-TW',
```
flodolo: These pages are not opt-in, and I assume the list of locales is always going to be the same for all of them?

If that's the case, to make this less error-prone, you can also define an array at the top of the file, and only reference the array here ([example](https://github.com/mozilla-l10n/langchecker/blob/master/app/config/sources.inc.php#L29)).

peiying2: That's true, the same locales for all three pages

craigcook: Updated!

app/config/sources.inc.php
```diff
@@ -407,14 +412,13 @@
...
-        ],
+        'supported_locales' => $participation_locales,
     ],
     'mozorg/about/governance/policies/reporting.lang' => [
-        'supported_locales' => [
-            'de', 'es-ES', 'fr', 'hi-IN', 'ja', 'pt-BR', 'zh-TW',
-        ],
+        'supported_locales' => $participation_locales,
+    ],
+    'mozorg/about/governance/policies/community-hotline.lang' => [
```
peiying2: nit pick: let's make sure it is in alphabetic order.

PR #960, opened 2020-02-21
Add browser comparison pages

craigcook:
Five new pages comparing Firefox to other major browsers, as well as a shared lang file.

We'll be handling French and German in-house. This is a LOT of text so all other locales are lowest priority, no deadline, and purely opt-in so localizers should only work on these pages if they want to.

Lang files added in https://github.com/mozilla-l10n/www.mozilla.org/pull/371

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -3,6 +3,7 @@
 $no_active_tag = [
     'download_button.lang',
     'firefox/shared.lang',
+    'firefox/compare/shared.lang',
     'footer.lang',
     'main.lang',
     'mozorg/internet-health/shared.lang',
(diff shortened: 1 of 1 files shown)


craigcook:
D'oh, looks like I can't add the shared.lang to the no_active_tag list until it actually exists. Should I remove it and add it later, or we can just rerun the Travis build after the other PR is merged?

flodolo:
> D'oh, looks like I can't add the shared.lang to the no_active_tag list until it actually exists. Should I remove it and add it later, or we can just rerun the Travis build after the other PR is merged?

You added a file to the no_active_tag, but you didn't define it in the list of supported files for mozilla.org

craigcook:
🤦‍♂

app/config/sources.inc.php
```diff
@@ -133,6 +134,48 @@
         ],
         'supported_locales' => $mozillaorg,
     ],
+    'firefox/compare/shared.lang' => [
```
peiying2: Nit pick: this goes after the new Safari entry.

craigcook:
Reordered and squashed

PR #963, opened 2020-02-28
Add whatsnew_74.lang

craigcook:
Lang file added in https://github.com/mozilla-l10n/www.mozilla.org/pull/377

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -393,6 +393,14 @@ $mozillaorg_lang = [
         ],
         'supported_locales' => $firefox_locales,
     ],
+    'firefox/whatsnew_74.lang' => [
+        'deadline'          => '2020-03-05',
+        'priority'          => [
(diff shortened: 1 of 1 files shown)


peiying2:
Looks good but I don't know if the communities have time to do anything. According to the release manager, the deadline for locale list is Feb 28. Release date is March 10th. How accurate is the deadline date?

craigcook:
> How accurate is the deadline date?

Not accurate at all, I just picked a date a few days before the release. We're doing DE and FR in-house so those are the only ones we can guarantee will be done by the release date, then it's just whoever else can finish in time.

If we use the current active list for whatsnew_73.lang, anyone who can finish 74 before the release can still get the 74 page, or else fall back to the 73 page. Will that work?

peiying2:
> If we use the current active list for whatsnew_73.lang, anyone who can finish 74 before the release can still get the 74 page, or else fall back to the 73 page. Will that work?

Yes, I will communicate it in my email. let's make sure it falls back on WNP73. Thx

PR #969, opened 2020-03-24
Add page 6 and 7

Errolyn:
strings here: https://github.com/mozilla-l10n/www.mozilla.org/pull/382/files
code here: https://github.com/mozilla/bedrock/pull/8655

Note: Eric will be adding French and German translations


Errolyn:
Sounds good!

PR #976, opened 2020-04-08
add enterprise

Errolyn:
strings here: https://github.com/mozilla-l10n/www.mozilla.org/pull/387


Errolyn:
@peiying2  I believe so

PR #978, opened 2020-04-22
add whatsnew-76

Errolyn:
Bedrock PR: https://github.com/mozilla/bedrock/pull/8782
L10N repo: https://github.com/mozilla-l10n/www.mozilla.org/pull/390

URL: https://www.mozilla.org/firefox/76.0/whatsnew/all/

### DIFF
--- a/app/config/sources.inc.php
+++ b/app/config/sources.inc.php
@@ -409,6 +409,14 @@ $mozillaorg_lang = [
         ],
         'supported_locales' => $firefox_locales,
     ],
+    'firefox/whatsnew_76.lang' => [
+        'deadline'          => '2020-04-30',
+        'priority'          => [
+            1 => ['de', 'fr', 'en-GB', 'en-CA'],
+            3 => ['all'],
+        ],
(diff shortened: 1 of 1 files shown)


app/config/sources.inc.php
```diff
@@ -409,6 +409,14 @@
         ],
         'supported_locales' => $firefox_locales,
     ],
+    'firefox/whatsnew_76.lang' => [
+        'deadline'          => '2020-04-30',
```
peiying2, line 413: The deadline should be 4/24 this Friday. The completed locale list is due to releng 10 days before the launch. Please change.

peiying2: Could you confirm what the fallback page is if WNP76 is incomplete for non-P1 locales?

Errolyn: It would be our standard fallback page: https://www.mozilla.org/en-US/firefox/whatsnew/all/

flodolo:
The build is green on Travis, but for some reason the result doesn't get back to GH.

PR #1009, opened 2020-12-19
Switch from Travis to GitHub Actions

flodolo:


flodolo:
Why in the world is Travis still trying to run? First case in 5 repos I've migrated today 🙄

PR #1015, opened 2021-04-29
Drop mozilla.org from langchecker

flodolo:
CC @peiying2 

Théo, can you double check that everything works for you? I tested locally, and nothing seems to explode.

The next step would be removing the running instance from https://l10n.mozilla-community.org/


TheoChevalier (approved):
Looking good!

